PR #9, opened 2018-04-23
Allow stable images to be bulk copied from one stream

smarterclayton:
Rather than pulling one tag from 40 image streams, we can use 1 stream
to represent the 40 components of a release and fetch those images in
bulk. Two new fields are added to the ci-operator api:

1. For the release tag step, the `name` field indicates the operator
should use a single image stream by name and copy all of its tags, and
then when the build is complete, any images should be published to that
tag
2. For the image step, a new `as` field on an image stream tag reference
allows the caller to define the name of the component on the pipeline.
Used to allow the caller to choose between the tag or stream name as the
"component name".


smarterclayton:
@stevekuznetsov 

Tested with:

```
$ oc tag ci/release-with-clonerefs:golang-1.9 release:golang-1.9
$ JOB_SPEC="$( cat test/env )" ./ci-operator --build-config="$( cat test/config.json )" --namespace=test-ci-op --base-namespace=test-ci-op --dry-run=false
```

on the api server.

PR #10, opened 2018-04-25
Support --write-params to output dynamic parameter info

smarterclayton:
Write NAMESPACE, IMAGE_FORMAT, and RPM_REPO to an env formatted file
when --write-params=PATH is provided. Allows automation to use these
variables for later behavior.

```
$ ci-operator ... --write-params=/tmp/out
$ cat /tmp/out
NAMESPACE="ci-op-b11b8c1ba3"
IMAGE_FORMAT="registry.svc.ci.openshift.org/openshift/origin-v3.10:${component}"
RPM_REPO="rpm-repo-ci-op-b11b8c1ba3.svc.ci.openshift.org"
```

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -4,18 +4,25 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
 	"strings"
+	"time"
+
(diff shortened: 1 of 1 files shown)


smarterclayton:
@stevekuznetsov

smarterclayton:
Pushed a follow up, IMAGE_FORMAT needs to be single-quoted thanks to bash

PR #11, opened 2018-04-30
As a convenience for a test, write out the JOB_NAME as a param

smarterclayton:
Params are expected to be passed to other elements, and JOB_NAME is the
unique identifier for type within a PR.


smarterclayton:
@stevekuznetsov used similarly to $NAMESPACE, to disambiguate actions by the code calling the ci-operator.

PR #16, opened 2018-05-02
Add template execution and secret creation to the operator

smarterclayton:
Templates are passed to the command, parsed, and then used as arbitrary
execution units in the CI graph. The operator requires that each
template contain at least one restart=Never pod and will set up
dependencies on earlier stages based on the parameters passed as input.
The following parameters are automatic:

* JOB_NAME: the name of the job from the Job spec
* JOB_NAME_HASH: a short hash of the job name
* JOB_NAME_SAFE: the job name converted to a Kubernetes safe name
* NAMESPACE: the target namespace
* IMAGE_FORMAT: requires the release stage be defined
* RPM_REPO: a URL pointing to the hosted RPM repo
* IMAGE_*: depends on the release stage and places the wildcard value
    into IMAGE_FORMAT as ${component}

Environment variables can underfill the job.

The pods in each template are considered successful when they reach
phase=Succeeded and considered failed if any container or init container
has a non-zero exit code. The template name is used as the unique key.

Also support `--secret-dir` which converts a directory into a secret with the same name as the directory.


smarterclayton:
Added a test step command.  Fixed a bug where multiple identical Requires links caused multiple executions of the step.  Moved from `--build-config JSON` to taking `CONFIG_SPEC` as environment.

smarterclayton:
@stevekuznetsov

smarterclayton:
Added `--target` to restrict the graph to only steps that are dependencies of the source.

cmd/ci-operator/main.go
```diff
@@ -53,9 +53,23 @@ func main() {
...
+func (s *stringSlice) Set(value string) error {
+	s.values = append(s.values, value)
+	return nil
+}
+
 type options struct {
-	rawBuildConfig string
-	templatePaths  []string
+	rawBuildConfig    string
+	templatePaths     []string
```
stevekuznetsov: Why should we pass paths on the filesystem for the config, template and secrets instead of just having the in-cluster client that the ci-operator has read the data from the apiserver itself?

pkg/api/graph.go
```diff
@@ -107,7 +107,7 @@ func BuildGraph(steps []Step) []*StepNode {
 				for _, otherCreates := range other.Step.Creates() {
 					if nodeRequires.Matches(otherCreates) {
 						isRoot = false
-						other.Children = append(other.Children, node)
```
stevekuznetsov: How did you end up with nodes in `allNodes` more than once?

pkg/api/graph_test.go
```diff
@@ -143,6 +143,18 @@ func TestBuildGraph(t *testing.T) {
 		creates:  []StepLink{InternalImageLink(PipelineImageStreamTagReference("final"))},
 	}
 
+	duplicateRoot := &fakeStep{
+		requires: []StepLink{ExternalImageLink(ImageStreamTagReference{Namespace: "ns", Name: "base", Tag: "latest"})},
+		creates:  []StepLink{InternalImageLink(PipelineImageStreamTagReferenceRoot)},
+	}
+	duplicateSrc := &fakeStep{
+		requires: []StepLink{
+			InternalImageLink(PipelineImageStreamTagReferenceRoot),
```
stevekuznetsov: Oh ew we should just disallow this config?

pkg/steps/rpm_server.go
```diff
@@ -245,7 +245,7 @@ func waitForDeploymentOrTimeout(client appsclientset.DeploymentConfigInterface,
 				return false, nil
 			}
 			if deploymentFailed(deploymentConfig) {
-				return false, fmt.Errorf("the DeploymentConfig %s/%s failed with status %q", deploymentConfig.Namespace, deploymentConfig.Name, deploymentReason(deploymentConfig))
+				return false, fmt.Errorf("the deployment config %s/%s failed with status %q", deploymentConfig.Namespace, deploymentConfig.Name, deploymentReason(deploymentConfig))
```
stevekuznetsov: meh

cmd/ci-operator/main.go
```diff
@@ -91,22 +90,22 @@ func bindOptions() *options {
...
 func (o *options) Validate() error {
-	if o.rawBuildConfig == "" {
-		return fmt.Errorf("job configuration must be provided with `--build-config`")
-	}
 	return nil
 }
 
 func (o *options) Complete() error {
-	if err := json.Unmarshal([]byte(o.rawBuildConfig), &o.buildConfig); err != nil {
+	configSpec := os.Getenv("CONFIG_SPEC")
```
stevekuznetsov: `os.LookupEnv`?

pkg/api/types.go
```diff
@@ -139,6 +142,21 @@ type PipelineImageCacheStepConfiguration struct {
...
 
+type TestStepConfiguration struct {
+	// As is the name of the test.
+	As string `json:"as"`
+	// From is the image stream tag in the pipeline to run this
+	// command in.
+	From PipelineImageStreamTagReference `json:"from"`
+	// Commands are the shell commands to run in
+	// the repository root to execute tests.
+	Commands string `json:"commands"`
```
stevekuznetsov: Why is it on us to do word splitting?

pkg/steps/test.go
```diff
@@ -0,0 +1,100 @@
...
+		ObjectMeta: meta.ObjectMeta{
+			Name: s.config.As,
+		},
+		Spec: coreapi.PodSpec{
+			RestartPolicy: coreapi.RestartPolicyNever,
+			Containers: []coreapi.Container{
+				{
+					Name:    "test",
+					Image:   fmt.Sprintf("%s:%s", PipelineImageStream, s.config.From),
+					Command: []string{"/bin/bash", "-c", "#!/bin/bash\nset -euo pipefail\n" + s.config.Commands},
```
stevekuznetsov: I am weary of this

cmd/ci-operator/main.go
```diff
@@ -87,6 +89,7 @@ type options struct {
 
 func bindOptions() *options {
 	opt := &options{}
+	flag.Var(&opt.targets, "target", "A set of names in the config to target. Only steps that are required for these targets will be run.")
```
stevekuznetsov: What's the use-case for more than one leaf here?

pkg/steps/pipeline_image_cache.go
```diff
@@ -48,6 +48,8 @@ func (s *pipelineImageCacheStep) Provides() (api.ParameterMap, api.StepLink) {
 	return nil, nil
 }
 
+func (s *pipelineImageCacheStep) Name() string { return string(s.config.To) }
```
stevekuznetsov: Why do we want any steps other than leaves to be targeted?

pkg/steps/source.go
```diff
@@ -227,8 +227,7 @@ func buildDuration(build *buildapi.Build) time.Duration {
 
 func printBuildLogs(buildClient BuildClient, name string) {
 	if s, err := buildClient.Logs(name, &buildapi.BuildLogOptions{
-		NoWait:     true,
-		Timestamps: true,
```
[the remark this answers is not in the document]
smarterclayton: Timestamps were confusing and ragged.  Wait was an accident.

pkg/steps/template.go
```diff
@@ -0,0 +1,492 @@
...
+			if len(value) > 0 {
+				s.template.Parameters[i].Value = value
+			}
+			continue
+		}
+		if strings.HasPrefix(p.Name, "IMAGE_") {
+			component := strings.ToLower(strings.TrimPrefix(p.Name, "IMAGE_"))
+			if len(component) > 0 {
+				component = strings.Replace(component, "_", "-", -1)
+				format, err := s.params.Get("IMAGE_FORMAT")
```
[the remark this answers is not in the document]
smarterclayton: why?

pkg/api/graph.go
```diff
@@ -8,8 +8,11 @@ type Step interface {
 
 	Requires() []StepLink
 	Creates() []StepLink
+	Provides() (ParameterMap, StepLink)
```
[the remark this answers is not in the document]
smarterclayton: The step link is only used if you use the parameter.

cmd/ci-operator/main.go
smarterclayton: Because then you can't run it locally.

smarterclayton: Data is read from one place and written to one place.  We read from the filesystem and write to the server.  That's why pods have injection points - dynamic reading is unnecessary ceremony.

pkg/api/graph.go
smarterclayton: A step link parameter that depended on release tags and a regular Requires() that depended on release tags.

pkg/api/graph_test.go
smarterclayton: Provides() is a dynamic dependency.  We could make the step links resolve themselves, but it's a pretty obvious error to make.

pkg/steps/rpm_server.go
smarterclayton: kinds are not things users deal with.

smarterclayton:
Addressed or commented, changes pushed.

smarterclayton:
Added a `-h` and provided help.

smarterclayton:
Made the graph include the images we will tag in for prereqs part of the input hash, which required two changes:

1. a pass over all steps prior to building the graph to calculate all their inputs (the steps cache the resolution)
2. the namespace that steps create into must be lazily resolved, so switch all the clients passed into steps from *Interface to *Getter.  

I debated whether to change the signature of Run and decided against it because it keeps the run action independent of the actual graph being executed.  Steps are already lazily evaluating state from the source of truth (the cluster) so the Run() command just asks the JobSpec what namespace it should run from.  We don't attach things to context generally.

After this commit, builds are hermetic which is a nice win (two different users with the same source, build definition, and base images will get the same output).  This was required to make retries work:

1. test job A runs the first time while a particular dependency (base image) is broken
2. test job A fails because the base image is broken
3. user retries test job A as job B, but if job B won't rebuild the artifacts we'll remain broken

After this change since the base image is changed, the input hash changes, which means a new namespace is used.

pkg/steps/template.go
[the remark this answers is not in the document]
stevekuznetsov: I'm dumb and will make typos?

pkg/api/graph.go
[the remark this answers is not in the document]
stevekuznetsov: I'm 10,000% unclear on this API still. When is the `StepLink` from this `Provides()` different from that from `Creates()`?

pkg/api/types.go
stevekuznetsov: Isn't it? We take this single string and need to feed it into a `[]string{}` for the Pod, so isn't every entry in that array treated as a single arg?

cmd/ci-operator/main.go
stevekuznetsov: When you are building images, just run the build job to completion. When you need a leaf, you are running a single test. If the test requies multiple parents, make the tree relationship as necessary? Using `--target` for non-test runs seems wrong.

smarterclayton:
@stevekuznetsov I get lonely without your comments... :(

pkg/steps/template.go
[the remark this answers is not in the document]
smarterclayton: If you make typos things will break.  I prefer to not overuse constants for things that shouldn’t ever change.  Adding a constant is adding a pointer lookup for a human so you can actually make your code harder to read.  

Also, if you take a code dependency on an implementation to get a constant that’s bad.  This is part of an api, you can’t change the constant anyway safely.

Within a package I think it’s ok but often overkill.

pkg/api/types.go
smarterclayton: We’re feeding a scriptlet to bash as arg[2].  The same as the build command.

smarterclayton:
I’ll make target singular.  I can unify creates and provides in a follow up?

smarterclayton:
Disabled multiple targets.  Will come back to creates and provides.

PR #19, opened 2018-05-17
Gather artifacts from test pods and template pods

smarterclayton:
A test pod must have "artifact_dir" defined and the user must pass a top
level --artifact-dir flag to gather artifacts. A template pod will gather
artifacts if the flag is set and the pod has a volume called
"artifacts" and that volume is mounted into any of the pod containers.

Artifacts are gathered by invoking a tar exec session into a busybox
sidecar. After all containers have gathered logs a file is removed in the
container via exec which cases the session to exit.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -92,6 +92,12 @@ of dynamic parameters that are inferred from previous steps. These parameters ar
 
 Dynamic environment variables are overriden by process environment variables.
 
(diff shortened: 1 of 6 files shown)


smarterclayton:
@kargakis what's our canonical GCS artifact upload pattern for a prow job?  do i need to add a sidecar?  add to a directory?

smarterclayton:
Ok, going to poke that

PR #20, opened 2018-05-20
Allow override of the default config via --override

### LINKED ISSUE #17: Have canonical inputs to config.json that can be overriden
config.json takes a number of input images (that today must come from a cluster)

```
{
  "tag_specification": {
    "namespace": "openshift",
    "name": "origin-v3.10",
    "tag": "",
    "tag_overrides": {}
  },
  "base_rpm_images": [
    {
      "namespace": "openshift",
      "name": "origin-v3.10",
      "tag": "base",
      "as": "base"
    }
  ],
  "test_base_image": {
    "namespace": "ci",
    "name": "release-with-clonerefs",
    "tag": "golang-1.9"
  },
```

These are all details of the particular cluster a job is run against, not a detail about the project itself.  each of these steps results in something being tagged into either `stable` or `pipeline` with a short name (which must be unique).

I think the canonical form of these should either be in the dockerfiles themselves or as regular docker image references (`docker.io/openshift/origin-base:v3.10`) where the CI system can override them by name (`use namespace openshift name origin-v3.10 as 'base'`) and a local developer running CI operator against an arbitrary cluster can see it simply work.

Suggested change:

1. change `base_rpm_images` and `test_base_image` to be something a bit more generic:

```
{
  "images": {
    "test-base": {"image": "docker.io/openshift/origin-release:golang-1.9"},
    "base": {"image": "docker.io/openshift/origin-base:latest"} 
  },
  "tag_specification": {
     "cluster": "api.ci.openshift.org",
    "namespace": "openshift",
    "name": "origin-v3.10"
  },
}
```

which can also be overridden via command line flag `--images`:

```
... --images="$( cat images.json )"
$ cat images.json
{
  "images": {
    "test-base": {"namespace": "ci", "name": "release-with-clonerefs", "tag": "golang-1.9"},
    "base": {"namespace": "openshift", "origin-v3.10", "tag": "base"},
  },
  "tag_specification": {
    "namespace": "openshift",
    "name": "origin-v3.10",
    "tag": "",
    "tag_overrides": {}
  },
}
```

The CI system would selective override images and tag specification.  The end user would be able to run ci-operator locally.

smarterclayton:
Fixes #17

Allows us to separate out CI job definitions provided with the repo from the inputs we want to use.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -104,6 +104,9 @@ that defines artifact_dir or template that has an "artifacts" volume mounted
 into a container will have artifacts extracted after the container has completed.
 Errors in artifact extraction will not cause build failures.
 
+In CI environments the inputs to a job may be different than what a normal
+development workflow would use. The --override file will override fields
+defined in the config file, such as base images and the release tag configuration.
 `
 
 func main() {
@@ -151,6 +154,7 @@ func (s *stringSlice) Set(value string) error {
 
 type options struct {
 	configSpecPath    string
+	overrideSpecPath  string
 	templatePaths     stringSlice
 	secretDirectories stringSlice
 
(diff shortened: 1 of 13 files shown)


smarterclayton:
Three separate changes:

1. `--override` and `OVERRIDE_SPEC` can be layered on top of a repo-provided spec - this allows the repo provided spec to be tailored for a CI environment
2. Allow images to be referenced from a cluster anonymously, so the repo can be used by an unauthenticated user (api.ci.openshift.org has anonymous read access to openshift/origin-X)
3. Allow `resources` to be specified as a top level field in the config, with a `*` default request/limit and a per step (`as` or the image name) value that is placed on pods and builds.

PR #22, opened 2018-05-29
Add a promotion definition that will copy pipeline tags out

smarterclayton:
Branch jobs should be able to promote the known tags to a target
location when needed. The alternative is to duplicate the list of images
known in the config, which complicates setup. --promote is an optional
flag that instructs the ci-operator to also attempt a promotion. This
prevents the job from being able to automatically promote (when in a
repo config). The promotion step is executed after the graph and
promotes anything that was successfully built.

@kargakis @stevekuznetsov last step to have a branch job publish images back to the converged list

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -158,7 +158,8 @@ type options struct {
 	templatePaths     stringSlice
 	secretDirectories stringSlice
 
-	target string
+	target  string
+	promote bool
 
 	verbose bool
 	help    bool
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
Publishing to external registries and publishing of non-image data are pretty out there -- definitely useful features but perhaps not in the scope of this codebase? Could we run a step that does the publish using utilities in the container image?

cmd/ci-operator/main.go
```diff
@@ -362,7 +366,17 @@ func (o *options) Run() error {
 		}
 
 		// execute the graph
-		return steps.Run(ctx, nodes, o.dry)
+		if err := steps.Run(ctx, nodes, o.dry); err != nil {
+			return err
+		}
+
+		for _, step := range postSteps {
```
stevekuznetsov: This is scary -- why are we not reinventing Jenkins here?

PR #25, opened 2018-06-15
Inject clonerefs into the source builds

stevekuznetsov:
We do not want users to have to inject our infrastructure tools into
their test containers, so we can do it ourselves before we need to use
them.

/assign @smarterclayton 
This is going to add a bunch of boilerplate I guess

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -289,10 +289,15 @@ type SourceStepConfiguration struct {
 	From PipelineImageStreamTagReference `json:"from"`
 	To   PipelineImageStreamTagReference `json:"to,omitempty"`
 
-	// SourcePath is the location within the source repository
+	// PathAlias is the location within the source repository
 	// to place source contents. It defaults to
 	// github.com/ORG/REPO.
 	PathAlias string `json:"source_path"`
+	// ClonerefsImage is the image where we get the clonerefs tool
+	ClonerefsImage ImageStreamTagReference `json:"clonerefs_image"`
+	// ClonerefsPath is the path in the above image where the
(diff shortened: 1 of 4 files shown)


pkg/api/types.go
```diff
@@ -38,6 +38,10 @@ type ReleaseBuildConfiguration struct {
 	// cloning from is ignored.
 	CanonicalGoRepository string `json:"canonical_go_repository"`
 
+	// ClonerefsSourceImage configures from which image we extract the clonerefs
+	// tool for use in our clone steps
+	ClonerefsSourceImage ClonerefsSourceImage `json:"clonerefs_source_image"`
```
smarterclayton: I would call this `release_tools`.

smarterclayton: or something that is more generic.

stevekuznetsov: ack

pkg/api/types.go
```diff
@@ -66,6 +69,19 @@ type ReleaseBuildConfiguration struct {
 	Resources ResourceConfiguration
 }
 
+// ReleaseToolsConfig holds configuration for tools used in the infrastructure
+type ReleaseToolsConfig struct {
+	// ClonerefsSourceImage configures from which image we extract the clonerefs
+	// tool for use in our clone steps
+	ClonerefsSourceImage ClonerefsSourceImage `json:"clonerefs_source_image"`
```
smarterclayton: Are we really, really, really going to want to have multiple of these?  Why wouldn't we just have a release tools image.

spadgett:
I didn't know I needed a /yolo command until just now

PR #27, opened 2018-06-21
Support additional image promotions during promote

smarterclayton:
The user can add an additional promotion step for images that aren't in the
images list (so we can promote a base image, or an input image, or rename
an image into multiple spots).

Used by origin to push origin-base.

```
  "promotion": {
    "namespace": "ci",
    "name": "test-stable",
    "additional_images": {
      "base": "base",
      "rpms": "rpms",
      "src": "src"
    }
  },

```


smarterclayton:
@stevekuznetsov can you take a quick look?  builds, fixed a test break, and verified it locally for origin (changes in a separate PR to the prow config)

PR #29, opened 2018-06-22
Perform a submodule init on the cloned code

smarterclayton:
Not currently optional, but could be made so. Required for
openshift/origin-aggregated-logging

@stevekuznetsov if we make this optional i think it should be a boolean on the top level - but i'm not sure what scenario you have a git submodule in your repo that you don't need it for normal workloads.


smarterclayton:
Let me know your preference.

stevekuznetsov:
Why shouldn't this be in `clonerefs`?

PR #35, opened 2018-06-25
Use a multithreaded python server to serve RPMs

smarterclayton:
SimpleHTTPServer is single threaded and so a long download will cause
liveness checks to fail. This is a more complicated variant that runs
multiple threads.

@stevekuznetsov 

Tested manually, verified shutdown is clean, ran a load test against it.


stevekuznetsov:
I wish I never saw this :O

PR #36, opened 2018-06-25
Expose 'src' as LOCAL_IMAGE_SRC for use in template jobs

smarterclayton:


smarterclayton:
Other jobs create this binding.

PR #49, opened 2018-07-16
Fix error messages on client creation failures

petr-muller:
Noticed these while reading the code, they look like c/p artifacts.


petr-muller:
@stevekuznetsov ptal

stevekuznetsov:
Merge queue was just stuck :)

PR #57, opened 2018-07-24
Add README, onboarding and config reference docs

petr-muller:
First shot at the documentation for CI Operator, mostly to gather feedback about the structure and content. The intent is to:

1. Provide `README.md` with basic information and links
2. Provide `ONBOARD.md` to walk through how to configure basic jobs for a OpenShift component repository
3. ~~Provide `CONFIG.md` to describe `config.json` keys~~ (deferred for later, see #70)


ONBOARD.md
```diff
@@ -0,0 +1,104 @@
+## Prepare configuration for component repo
+
+The JSON configuration file describes how to build different images in a
+testing pipeline for your repository. CI Operator has different *”targets”*
+that describe the goal images to build, and later targets build on successfully
+built earlier targets.
+
+### Source code image target
+
+By default, CI Operator builds the `src` target image. The `src` image is
```
smarterclayton: Use `ci-operator` everywhere instead of `CI Operator`.  It's a binary tool vs a more formal project.

ONBOARD.md
```diff
@@ -0,0 +1,104 @@
+## Prepare configuration for component repo
+
+The JSON configuration file describes how to build different images in a
```
smarterclayton: Add a few sentences to this paragraph describing how these are instructions for adding new CI jobs to the OpenShift CI systems.

README.md
```diff
@@ -0,0 +1,45 @@
+# CI Operator
```
smarterclayton: ci-operator

README.md
```diff
@@ -0,0 +1,45 @@
...
+file, describing base images and which images should be built and tested and
+how, CI Operator builds the component images within an OpenShift cluster and
+runs the tests. All artifacts are built in a new namespace named using a hash of
+all inputs, so the artifacts can be reused when the inputs are identical.
+
+CI Operator is mainly intended to be run inside a pod in a cluster, triggered by
+the Prow CI infrastructure, but it is also possible to run it as a CLI tool on a
+developer laptop.
+
+Note: CI Operator is not a [Kubernetes operator](https://coreos.com/operators/).
```
smarterclayton: I think this reads wierd - you might want to say "ci-operator orchestrates builds and tests, but should not be confused with [Kubernetes operators](....) which make managing software on top of Kubernetes easier."

petr-muller:
Addressed. It's sausage: I'll squash before the eventual merge or course.

petr-muller:
One thing I'm fiddling with is a `config.json` reference, started in the `CONFIG.md` file. I expected to create a table like this:

| Key | Description | `example` |
| --- | --- | --- |

but `example` would usually be a multi-line snippet which Markdown table cannot do (or at least I do not how). It would be possible to achieve this with inline HTML but I am not sure I want to go there.

petr-muller:
Yeah, types.go already have descriptions, but the connection to the actual keys that people need to write to the config would not be exactly clear...

I'm inclined to just create a heavily annotated example file with all the possible keys in it. JSON does not have comments so it would not be a real config but that is fine IMO.

CONFIG.md
```diff
@@ -0,0 +1,18 @@
+# ci-operator configuration file
```
stevekuznetsov: Can we use the upstream codegen to generate this file from the structs?

ONBOARD.md
```diff
@@ -0,0 +1,113 @@
...
+testing pipeline for your repository. ci-operator has different *”targets”*
+that describe the goal images to build, and later targets build on successfully
+built earlier targets.
+
+### Source code image target
+
+By default, ci-operator builds the `src` target image. The `src` image is
+expected to contain the component source code together with all build
```
stevekuznetsov: You could mention that the code that is cloned will correctly contain the refs under test, whether that is a branch or a PR or a batch. You could also link to the `clonerefs` tool: https://github.com/kubernetes/test-infra/tree/master/prow/cmd/clonerefs

ONBOARD.md
```diff
@@ -0,0 +1,113 @@
...
+dependencies. You can specify the base image to which the source code will be
+injected with the `test_base_image` key (the base image will almost always be
+some `openshift/release:<tag>` image.
+
+```
```
stevekuznetsov: Could remove the `cat` line here and use highlighting:

    ```json

ONBOARD.md
```diff
@@ -0,0 +1,113 @@
...
+    "tag": "golang-1.10"
+  }
+}
+```
+
+Given your component can be built in the context of the `openshift/release`
+image, you can test building the `src` target:
+
+```
+$ ./ci-operator --config example.json --namespace 'ciop-test' --git-ref=openshift/<component>@<revision> --target=src
```
stevekuznetsov: Do you need `--namespace` here? Won't we make one as necessary?

ONBOARD.md
```diff
@@ -0,0 +1,113 @@
...
+}
+```
+
+### Intermediary binary targets
+
+Two test targets in the previous example assume their `make` targets take care
+of full build from source till the actual test. This is often the case, but it
+is ineffient because each test target will then perform the build separately. CI
+operator can create `bin` and `test-bin` targets for the test targets to share
+by providing `binary_build_commands` and `test_binary_build_commands`
```
stevekuznetsov: We have two test lists here because often it's a different compilation process for test binaries than for normal ones -- in Go that is the difference between "normal" compilation and compilation to test race conditions

README.md
```diff
@@ -0,0 +1,47 @@
+# ci-operator
+
+ci-operator automates and simplifies the process of building and testing
+OpenShift component images (i.e. any `openshift/origin-{component}` images).
+
+Given a Git repository reference and a component-specific JSON configuration
```
stevekuznetsov: don't call-out JSON as we will change that in the future

ONBOARD.md
```diff
@@ -0,0 +1,113 @@
+This document describes how to add new CI jobs (utilizing ci-operator) for
+OpenShift components to the OpenShift CI system. In general, two steps are
+needed:
+
+1. [Prepare configuration file](#prepare-configuration-for-component-repo),
```
sttts: where is this config file supposed to be?

ingvagabund:
@paulfantom PTAL

ONBOARD.md
stevekuznetsov: openshift/release/ci-operator/...

sttts: How about writing that here in the doc?

petr-muller:
Feedback to be addressed:

- [ ] README.md probably needs a basic "Usage" section: describe the environment to run it in, the CLI flags, and how to write a config file.
- [ ] Broken link in README.md: ONBOARDING.md --> ONBOARD.md

ONBOARD.md
petr-muller: Definitely! I will make the doc explicitly mention the "write config locally & test, then submit it to *there*" workflow and suggest the destination.

CONFIG.md
petr-muller: @smarterclayton also mentioned this. I guess I will try what comes out of the tool, but I am skeptical how well can struct documentation serve as a reference on how to write JSON.

README.md
petr-muller: IMO, once we do that, we can fix that in the doc. Now it's telling the truth. I will adjust the wording though to inform about the future.

CONFIG.md
smarterclayton: Even godoc is probably good enough.

smarterclayton: Actually, if you add an examples section https://godoc.org/github.com/openshift/ci-operator/pkg/api is pretty much correct.

petr-muller: I am not sure I would consider this very helpful if I wanted to write a config file... I mean yeah, the content is there if you know where to look, and it is correct, but the connection to how the config is written is *very* indirect if you are not familiar with how the thing works internally.

ONBOARD.md
```diff
@@ -0,0 +1,183 @@
...
+   describing how ci-operator should build and test your component.
+2. [Add new jobs to Prow](#add-prow-jobs), hooking ci-operator to important
+   GitHub events for your repository.
+
+This walkthrough covers a simple use case of hooking up a job running source
+code level tests (unit, performance etc.) on pull requests on a component
+repository. Although ci-operator can perform more sophisticated test steps than
+simply running unit tests, these use cases are beyond the scope of this doc.
```
stevekuznetsov: Bruno will add this in so maybe just note that the doc is coming

ONBOARD.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Prepare configuration for component repo
+
+The JSON configuration file describes how to build different images in a
+testing pipeline for your repository. ci-operator has different *”targets”*
```
stevekuznetsov: s/ci-operator/`ci-operator`/g

ONBOARD.md
```diff
@@ -0,0 +1,183 @@
...
+that describe the goal images to build, and later targets build on successfully
+built earlier targets. You will probably want to write and test your config file
```
stevekuznetsov: What does "test your config file" mean? How do you know it "works"?

ONBOARD.md
```diff
@@ -0,0 +1,183 @@
...
+of full build from source till the actual test. This is often the case, but it
+is ineffient because each test target performs the build separately. CI
+operator can create `bin` and `test-bin` targets for the test targets to share
+by providing `binary_build_commands` and `test_binary_build_commands`
+respectively (we have two test lists here because often it's a different
+compilation process for test binaries than for normal ones -- in Go that is the
+difference between "normal" compilation and compilation to test race
+conditions):
+
+```json
```
stevekuznetsov: This is getting tagged as invalid JSON -- maybe you are missing a comma somewhere?

ONBOARD.md
```diff
@@ -0,0 +1,183 @@
...
+execution to build just one or more targets using the `--target` option.
+
+### Intermediary binary targets
+
+Two test targets in the previous example assume their `make` targets take care
+of full build from source till the actual test. This is often the case, but it
+is ineffient because each test target performs the build separately. CI
+operator can create `bin` and `test-bin` targets for the test targets to share
+by providing `binary_build_commands` and `test_binary_build_commands`
+respectively (we have two test lists here because often it's a different
```
stevekuznetsov: I could copy-edit this later, but larger parentheticals like this could just be their own sentences.

ONBOARD.md
```diff
@@ -0,0 +1,183 @@
...
+    {
+      “as”: “performance”,
+      “from”: “test-bin”,
+      “commands”: “make test-performance”
+    }
+  ]
+}
+```
+
+Here, `unit` and `integration` targets will both be built from a `bin` image,
```
stevekuznetsov: Let's make them depend on `test-bin` since that is more likely in reality

ONBOARD.md
```diff
@@ -0,0 +1,183 @@
...
+which will be a result of running `make build` over a `src` image.
+
+### Submit the configuration file to openshift/release
```
stevekuznetsov: s|openshift/release|`openshift/release`|g

ONBOARD.md
```diff
@@ -0,0 +1,183 @@
...
+will need to add the file to the
+[openshift/release](https://github.com/openshift/release) repository,
+specifically to its `ci-operator/config/openshift` subdirectory
+[tree](https://github.com/openshift/release/tree/master/ci-operator/config/openshift).
+Each OpenShift component has a separate directory there, and there is a
+configuration file in it per branch.
+
+### Images targets, end-to-end tests and more
+
+Building the source code and running unit tests is the trivial basic use case.
```
stevekuznetsov: Shouldn't call this trivial -- not only is it pretty complicated, but if someone is struggling to self-service, we shouldn't make them feel bad by describing what they are struggling with as trivial.

Opting into `ci-operator` to build images and run a test or two is a huge win for the team, too -- they get quick test runs compared to VM testing, they get their images published to Docker Hub and Quay with everything else automatically when we do rebuilds, etc.

ONBOARD.md
```diff
@@ -0,0 +1,183 @@
...
+ci-operator is able to build component images, provision test clusters using
+them and run end-to-end tests on them. These use cases would use more features
+in both configuration file and Prow job and would not fit into this document.
```
stevekuznetsov: Just note that it will fit in later, as Bruno will likely edit this doc after you are done

stevekuznetsov: (also so that someone coming here to look for that information has some knowledge that it will be here later, rather than just knowing they've come to the wrong place without a pointer as to where else to look)

README.md
```diff
@@ -0,0 +1,57 @@
+# ci-operator
+
+ci-operator automates and simplifies the process of building and testing
+OpenShift component images (i.e. any `openshift/origin-{component}` images).
```
stevekuznetsov: We do not enforce that naming scheme so `e.g.` instead of `i.e.`

README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Given a Git repository reference and a component-specific configuration file
+(JSON at the moment, but we [plan to switch to YAML](#46)), describing base
+images and which images should be built and tested and how, ci-operator builds
+the component images within an OpenShift cluster and runs the tests. All
+artifacts are built in a new namespace named using a hash of all inputs, so the
+artifacts can be reused when the inputs are identical.
+
+ci-operator is mainly intended to be run inside a pod in a cluster, triggered by
```
stevekuznetsov: s/pod/`Pod`/g

petr-muller:
I would like to merge this even without the config file reference as initially intended, because it'll probably take more work (I'll file a separate issue for that). Content-wise, I think `README.md` and `ONBOARD.md` are sort-of ready (I need to address all the feedback ofc) and I would like to avoid adding more content especially to `ONBOARD.md`, because it's quite long already.

ONBOARD.md
petr-muller: I would rather not - it's not that helpful to someone who actually needs more sophisticated info, and it should not be long until Bruno merges that doc. Once he does, I'll put a link to it here.

ONBOARD.md
petr-muller: You sure? Well I can, but IMO we should distinguish ci-operator as a tool and ci-operator as an executable in the text. Using backticks everywhere will just add visual clutter IMO.

ONBOARD.md
petr-muller: It works once you can locally run `ci-operator` without `--target` and with `--git-ref` and it builds all targets successfully, provided the tests actually pass for that ref. I'll try to clarify.

ONBOARD.md
petr-muller: I tried to reword this so it reads better.

petr-muller:
I think all feedback was addressed. I have created #70 to track "we need config file reference doc".

PR #58, opened 2018-07-24
Use a job for idle cleanup so that pod eviction doesn't orphan namespaces

smarterclayton:
Pods can be evicted, which prevents them from running to completion. cleanup-when-idle depends on getting a chance to handle termination, but eviction will bypass that. Therefore we use a job instead of a single pod, and the job can retry up to 6 times (the default) if evicted.

Without this change, in contested resource scenarios namespaces are occasionally not cleaned up.


smarterclayton:
Tested cleanup manually.

PR #59, opened 2018-07-24
Don't tear down namespace when interrupted

smarterclayton:
Instead, we rely on the job to ensure we eventually detect idle. Means
we can safely be killed.


smarterclayton:
Follow up to #58, we now get evicted on scale down (good!) but we terminate namespace (bad!) so we only terminate on happy path, and rely on the job controller to retry us.

PR #61, opened 2018-07-25
When importing images from the same cluster, use refs

smarterclayton:
As image stream sizes get larger, it can take longer to import. When we
are importing from the same cluster use image stream image references
rather than forcing an import.


smarterclayton:
@bparees this works around the very long import we were seeing on 3.11 (in one direction).

PR #73, opened 2018-08-02
--git-ref: Fail fast when ref does not resolve

petr-muller:
Previously, `ci-operator` attempted to continue even when a ref passed
did not resolve to an existing commit, leading to later builds failing,
because cloneref did not fetch any code. This change makes `ci-operator`
to fail fast in this situation.

Resolves: openshift/release#1096

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -273,12 +273,19 @@ func (o *options) Complete() error {
 
 	jobSpec, err := steps.ResolveSpecFromEnv()
 	if err != nil {
-		spec, refErr := jobSpecFromGitRef(o.gitRef)
-		if refErr != nil {
+		if len(o.gitRef) == 0 {
+			// Failed to read $JOB_SPEC and --git-ref was not passed
 			return fmt.Errorf("failed to resolve job spec: %v", err)
+		} else {
+			// Failed to read $JOB_SPEC but --git-ref was passed, so try that instead
+			spec, refErr := jobSpecFromGitRef(o.gitRef)
+			if refErr != nil {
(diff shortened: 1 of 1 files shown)


petr-muller:
/assign @stevekuznetsov @smarterclayton 

@smarterclayton, you authored the error recovery attempt there, any thoughts?

cmd/ci-operator/main.go
```diff
@@ -752,7 +752,7 @@ func jobSpecFromGitRef(ref string) (*steps.JobSpec, error) {
 	out, err := exec.Command("git", "ls-remote", fmt.Sprintf("https://github.com/%s/%s.git", prefix[0], prefix[1]), parts[1]).Output()
 	sha := strings.Split(strings.Split(string(out), "\n")[0], "\t")[0]
 	if len(sha) == 0 || err != nil {
-		return &steps.JobSpec{Type: steps.PeriodicJob, Job: "dev", Refs: steps.Refs{Org: prefix[0], Repo: prefix[1], BaseSHA: parts[1]}}, nil
+		return nil, fmt.Errorf("ref '%s' does not point to any commit in '%s'", parts[1], parts[0])
```
stevekuznetsov: Can we surface the `err` here?

petr-muller: Good idea.

petr-muller:
Wait a bit, this is slightly tricky - there are different error conditions ($JOB_SPEC not being set, problematic `--git-ref` arguments) and their combination, and I do not want to show misleading error messages. I'll finish this tomorrow.

petr-muller:
Ready for re-review

PR #75, opened 2018-08-03
Use the NS TTL Controller instead of cleanup pod

stevekuznetsov:

/assign @smarterclayton 
/cc @petr-muller @bbguimaraes @droslean

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -18,12 +18,10 @@ import (
 	"strings"
 	"time"
 
-	batchapi "k8s.io/api/batch/v1"
 	coreapi "k8s.io/api/core/v1"
 	rbacapi "k8s.io/api/rbac/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
-	batchclientset "k8s.io/client-go/kubernetes/typed/batch/v1"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	rbacclientset "k8s.io/client-go/kubernetes/typed/rbac/v1"
 	"k8s.io/client-go/rest"
@@ -185,6 +183,7 @@ type options struct {
 	namespace           string
 	baseNamespace       string
 	idleCleanupDuration time.Duration
+	cleanupDuration     time.Duration
 
 	inputHash     string
 	secrets       []*coreapi.Secret
@@ -199,6 +198,7 @@ type options struct {
 func bindOptions(flag *flag.FlagSet) *options {
 	opt := &options{
 		idleCleanupDuration: time.Duration(1 * time.Hour),
+		cleanupDuration:     time.Duration(12 * time.Hour),
 	}
 
 	// command specific options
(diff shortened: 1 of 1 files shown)


cmd/ci-operator/main.go
```diff
@@ -599,110 +618,6 @@ func (o *options) initializeNamespace() error {
...
-							if [[ "${count}" -lt 1 ]]; then
-								count+=1
-								sleep ${WAIT} & wait
-								continue
-							fi
-							echo "No pods running for more than ${WAIT}s, deleting project ..."
-							oc delete project ${NAMESPACE}
-							exit 0
-						done
-						`,
```
bbguimaraes: RIP

stevekuznetsov:
Needs https://github.com/openshift/release/pull/1141

cmd/ci-operator/main.go
```diff
@@ -216,6 +213,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.StringVar(&opt.namespace, "namespace", "", "Namespace to create builds into, defaults to build_id from JOB_SPEC. If the string '{id}' is in this value it will be replaced with the build input hash.")
 	flag.StringVar(&opt.baseNamespace, "base-namespace", "stable", "Namespace to read builds from, defaults to stable.")
 	flag.DurationVar(&opt.idleCleanupDuration, "delete-when-idle", opt.idleCleanupDuration, "If no pod is running for longer than this interval, delete the namespace. Set to zero to retain the contents.")
+	flag.DurationVar(&opt.cleanupDuration, "delete-after", opt.cleanupDuration, "If namespace exists for longer than this interval, delete the namespace. Set to zero to retain the contents.")
```
bbguimaraes: Not entirely true now that this depends on a separate controller being present.

stevekuznetsov: What would you suggest>

bbguimaraes: I'd add "(requires openshift/ci-ns-ttl-controller)" or something.

You probably won't care if you're invoking on the command line or your own
cluster, but it's worth documenting.

stevekuznetsov: done

cmd/ci-operator/main.go
```diff
@@ -215,7 +212,8 @@ func bindOptions(flag *flag.FlagSet) *options {
 	// the target namespace and cleanup behavior
 	flag.StringVar(&opt.namespace, "namespace", "", "Namespace to create builds into, defaults to build_id from JOB_SPEC. If the string '{id}' is in this value it will be replaced with the build input hash.")
 	flag.StringVar(&opt.baseNamespace, "base-namespace", "stable", "Namespace to read builds from, defaults to stable.")
-	flag.DurationVar(&opt.idleCleanupDuration, "delete-when-idle", opt.idleCleanupDuration, "If no pod is running for longer than this interval, delete the namespace. Set to zero to retain the contents.")
+	flag.DurationVar(&opt.idleCleanupDuration, "delete-when-idle", opt.idleCleanupDuration, "If no pod is running for longer than this interval, delete the namespace. Set to zero to retain the contents. Requires the namespace TTL controller to be deployed.")
+	flag.DurationVar(&opt.cleanupDuration, "delete-after", opt.cleanupDuration, "If namespace exists for longer than this interval, delete the namespace. Set to zero to retain the contents. Requires the namespace TTL controller to be deployed.")
```
petr-muller: What happens if the TTL controller is *not* deployed? This is annotation-driven, so I assume that the namespace will just live forever, right? Can we detect it and error out or at least warn?

bbguimaraes: Correct, nothing would remove the namespace.

On the other hand, if it's not deployed, it's most likely a test cluster where
leaking namespaces is not a problem.

smarterclayton: The first case isn’t true.  It’s trivial easy to run without our infra, and we should keep it that way.  Doesn’t mean we need to support teardown when controller is not present, but don’t apply that philosophy to other parts.

smarterclayton:
This is flaking:

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/20587/pull-ci-origin-e2e-gcp/2265/

You need to either use patch or retry conflicts multiple times.

cmd/ci-operator/main.go
stevekuznetsov: Eh, using `ci-operator` to run your unit tests is trivial. Getting all of the templates, configs and secrets together to run a templatized test on your own _is_ hard, and we are still working through the best way of fixing that. Even Gabe gave up.

stevekuznetsov:
Hmm, will move to `Patch()`.

PR #76, opened 2018-08-06
Remove decoration in the prow job config example.

droslean:
Since the documentation is not focusing on explaining further the prow jobs, the decoration configuration in the prow job example is not needed.

### DIFF
--- a/ONBOARD.md
+++ b/ONBOARD.md
@@ -174,8 +174,6 @@ presubmits:
     rerun_command: "/test unit"
     always_run: true
     trigger: "((?m)^/test( all| unit),?(\\s+|$))"
-    decorate: true
-    skip_cloning: true
     spec:
       serviceAccountName: ci-operator
       containers:


droslean:
@petr-muller @stevekuznetsov @bbguimaraes 
Shal we remove the decoration at all or keep it and update our docs with the decoration configuration?

bbguimaraes:
This is intended for people new to Prow who are writing jobs for `api.ci`, so
I'd suggest keeping the configuration as it is.  Ideally, the generator we are
working on would hide these configuration details completely (i.e. all
ci-operator jobs should have `decorate: true` and `skip_cloning: true`, so
there's no need for a job author to know about them).

We can add a separate section to document the configuration necessary to run
jobs in a Prow instance in another cluster.

petr-muller:
I don't have a strong opinion - if having these is a sensible default, I would leave it as they are. If there is even a slight chance that someone does not want this, I think removing it is a good idea, as it makes the example shorter and less intimidating.

droslean:
Well, I agree that it's better to have a decoration but for the ONBOARD.md which is focusing on the first initial configuration, if the reader will follow this documentation, he won't be able to make it work because of the decoration.
So we have 2 options, remove it or keep it and update the documentation with the decoration configuration.

petr-muller:
If the current example leads the user to a setup which does not work without additional setup, then we should remove it from there.

stevekuznetsov:
I would either keep these _or_ make the example do something different - use the base `golang:1.10` image to run `go test ./...` or something, then remove these bits. We don't want to confuse someone who is trying to write a ci-operator test and does need the options

PR #79, opened 2018-08-08
add Makefile

paulfantom:
I noticed this repo doesn't have a Makefile, so here is one.

This makefile consists of 4 targets:
- `check` - runs `go lint` (only outside of container) and `go vet`
- `build` - create `bin` directory and build package
- `test` - run `go test ./...`
- `help` - autogenerated help with short stage explaination

As a bonus most of commands are run in `golang` container, so everything is independent of host environment. This can be turned off by clearing environment variable `DOCKER_CMD`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,27 @@
+.PHONY: all
+all: check test build
+
+DOCKER_CMD := docker run --rm -v "$(PWD)":/go/src/github.com/openshift/ci-operator:Z -w /go/src/github.com/openshift/ci-operator golang:1.10
+
+.PHONY: check
+check: ## Lint code
+	@echo -e "\033[32mRunning golint...\033[0m"
+#	go get -u github.com/golang/lint # TODO figure out how to install when there is no golint
+	golint ./cmd/...
+	@echo -e "\033[32mRunning go vet...\033[0m"
+	$(DOCKER_CMD) go vet ./cmd/...
+
+.PHONY: build
+build: ## Build binary
+	@echo -e "\033[32mBuilding package...\033[0m"
+	mkdir -p bin
+	$(DOCKER_CMD) go build -v -o bin/ci-operator cmd/ci-operator/main.go
+
+.PHONY: test
+test: ## Run tests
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,27 @@
+.PHONY: all
+all: check test build
+
+DOCKER_CMD := docker run --rm -v "$(PWD)":/go/src/github.com/openshift/ci-operator:Z -w /go/src/github.com/openshift/ci-operator golang:1.10
```
smarterclayton: Generally our makefiles don't make docker calls so that we can run them by default inside of containers (vs requiring everyone to have docker locally).  I.e. I have go on my Mac, don't need docker to run it.

paulfantom: That's why it is optional. Everything can be run without docker.

paulfantom: What if I inverted it? By default run in host env and when some variable is set, run in container. This way all your workflows would be preserved and as an addition you will have an option to have reporoducible build environment in containers.

stevekuznetsov:
Would love to see one that looks more like `ci-ns-ttl-controller` for build/test/lint -- no shell characters, no special docker support, just the basics: https://github.com/openshift/ci-ns-ttl-controller/blob/master/Makefile

We can use `hack/env` to containerize if we need to

paulfantom:
> We can use hack/env to containerize if we need to

@stevekuznetsov may I ask how? I can change Makefile to include only basics, but I would like to know your idea of how containerization could be applied with `hack/env`. It sounds interesting :smile:

stevekuznetsov:
See https://github.com/openshift/origin/blob/master/hack/env -- it's a wrapper that injects your repo into a container and runs whatever you passed, to give you a standard env for your run.

stevekuznetsov:
That approach allows for composition on top of any other script and keeps the container logic out of everything else

paulfantom:
@stevekuznetsov I have simplified Makefile by removing references to containerized workflow and most of shell characters (left only those in `help` target).

stevekuznetsov:
We should update the tests to call these `make` targets now

PR #82, opened 2018-08-10
Label all pods we create with the ProwJobID for artifacts

stevekuznetsov:
Our artifact gathering and pushing controller will push logs from
containers that terminate in Pods labeled with a ProwJobID. We need to
label our test with this so that we can get more logs pushed.

/cc @bbguimaraes @petr-muller @droslean @smarterclayton

### DIFF
--- a/pkg/steps/job_spec.go
+++ b/pkg/steps/job_spec.go
@@ -16,9 +16,10 @@ import (
 // we do not import it as importing test-infra is a
 // massive hassle.
 type JobSpec struct {
-	Type    ProwJobType `json:"type,omitempty"`
-	Job     string      `json:"job,omitempty"`
-	BuildId string      `json:"buildid,omitempty"`
+	Type      ProwJobType `json:"type,omitempty"`
+	Job       string      `json:"job,omitempty"`
+	BuildId   string      `json:"buildid,omitempty"`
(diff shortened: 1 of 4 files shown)


pkg/steps/job_spec.go
```diff
@@ -16,9 +16,10 @@ import (
 // we do not import it as importing test-infra is a
 // massive hassle.
 type JobSpec struct {
-	Type    ProwJobType `json:"type,omitempty"`
-	Job     string      `json:"job,omitempty"`
-	BuildId string      `json:"buildid,omitempty"`
+	Type      ProwJobType `json:"type,omitempty"`
+	Job       string      `json:"job,omitempty"`
+	BuildId   string      `json:"buildid,omitempty"`
+	ProwJobID string      `json:"prowjobid,omitempty"`
```
stevekuznetsov: Because we don't import we never got the updated spec, but Prow exposes this in the downward API

stevekuznetsov:
@bparees @sdodson once this merges and is deployed we should get logs from the provisioning and deprovisioning containers in our test for GCP

stevekuznetsov:
oh FFS we need it in the pod from the template ... ugh

stevekuznetsov:
@smarterclayton templates --

PR #83, opened 2018-08-10
Ignore the RPM repo when calculating soft TTLs

stevekuznetsov:

Follow-up to #75

### DIFF
--- a/pkg/steps/rpm_server.go
+++ b/pkg/steps/rpm_server.go
@@ -28,8 +28,9 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


smarterclayton:
The TTL check was only supposed to be for pods that can terminate (not restartalways).  We will add other services to the namespace over time, and the caller shouldn’t have to take that into account.  If you do that, this logic is unnecessary.

stevekuznetsov:
OK, we can update the ttl controller to look for restartalways as well

PR #85, opened 2018-08-11
Bind namespace admin to author, not cluster admin

stevekuznetsov:

/cc @smarterclayton


smarterclayton:
This isn't correct either - it gives the accessor use of our ci secrets in that target namespace.  You can't give admin.  You can only give view, or something a slight bit broader that doesn't allow secrets.

stevekuznetsov:
Commented on Slack, set of people who can possibly get access but are not in the RedHat company is like 3 and they are trusted, so moot point.

PR #86, opened 2018-08-11
Add more contexts to errors

stevekuznetsov:

This is most likely way overkill, but I hate seeing bare errors sitting there on the command line and having users ask me about them.

/cc @petr-muller @smarterclayton @droslean @bbguimaraes

### DIFF
--- a/pkg/steps/output_image_tag.go
+++ b/pkg/steps/output_image_tag.go
@@ -68,7 +68,7 @@ func (s *outputImageTagStep) Run(ctx context.Context, dry bool) error {
 		fmt.Printf("%s\n", istJSON)
 	} else {
 		if err := s.istClient.ImageStreamTags(toNamespace).Delete(ist.Name, nil); err != nil && !errors.IsNotFound(err) {
-			return err
+			return fmt.Errorf("could not remove output imagestreamtag: %v", err)
 		}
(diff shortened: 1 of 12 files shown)


smarterclayton:
Did you verify that none of the code that calls into these methods was checking the type of error at a higher level?

pkg/steps/output_image_tag.go
```diff
@@ -68,15 +68,15 @@ func (s *outputImageTagStep) Run(ctx context.Context, dry bool) error {
...
+			return fmt.Errorf("could not remove output imagestreamtag: %v", err)
 		}
 		_, err := s.istClient.ImageStreamTags(toNamespace).Create(ist)
 		if errors.IsAlreadyExists(err) {
 			// another job raced with us, but the end
 			// result will be the same so we don't care
 			return nil
 		}
-		return err
+		return fmt.Errorf("could not create output imagestreamtag: %v", err)
```
smarterclayton: This is incorrect, err can be nil.  I'll fix this instance, but I'll need you to verify the other cases aren't like this.

stevekuznetsov: Hmmmmm. I'll have to look through the rest tomorrow

PR #87, opened 2018-08-11
Retry namespace update for TTL labels

stevekuznetsov:

Still no canonical patch generation library, so this will do

/cc @smarterclayton

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -590,7 +590,16 @@ func (o *options) initializeNamespace() error {
 		for key, value := range updates {
 			update.ObjectMeta.Annotations[key] = value
 		}
-		if _, err := client.Namespaces().Update(update); err != nil {
+		var updateErr error
+		for retries := 5; retries > 0; retries-- {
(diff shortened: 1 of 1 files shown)


smarterclayton:
Should only be retrying on conflict.

smarterclayton:
and we have a helper for it that does this for you in the client library.

cmd/ci-operator/main.go
```diff
@@ -590,7 +590,16 @@ func (o *options) initializeNamespace() error {
 		for key, value := range updates {
 			update.ObjectMeta.Annotations[key] = value
 		}
-		if _, err := client.Namespaces().Update(update); err != nil {
+		var updateErr error
+		for retries := 5; retries > 0; retries-- {
+			if _, err := client.Namespaces().Update(update); err == nil {
```
smarterclayton: this won't work, you have to actually re fetch and reapply your changes.

stevekuznetsov: Oh ugh, we should just use patch. Why is that hard today with client libraries?

PR #98, opened 2018-08-21
Allow release tag configuration to be omitted for unrelated repos

smarterclayton:
ci-operator can be used without an opinionated release configuration (like origin)
and instead be used in "single project" mode. null release config should mean
"don't use output image or create stable"

Allows us to fix ci-operator.


stevekuznetsov:
Great changes, can we get doc on this ...

PR #100, opened 2018-08-22
Send events with some usefull information.

### LINKED ISSUE #68: ci-operator should write events when it completes and when it creates a namespace
So that the event stream contains a record.

@stevekuznetsov now that we are doing a better job at recording events.

droslean:
closes #68


droslean:
WIP to resolve vendors issue.

stevekuznetsov:
@droslean are you still struggling with the vendoring here?

smarterclayton:
I'm not sure why there are extra dependencies coming in.

cmd/ci-operator/main.go
```diff
@@ -814,3 +827,12 @@ func shorten(value string, l int) string {
 	}
 	return value
 }
+
+func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) record.EventRecorder {
```
smarterclayton: You should only create one of these and then pass it down to whatever we use.  Don't create it multiple times.

cmd/ci-operator/main.go
```diff
@@ -422,17 +423,26 @@ func (o *options) Run() error {
...
 		}
 		if err != nil {
+			eventRecorder(client, o.namespace).Event(&coreapi.ObjectReference{Namespace: o.namespace}, coreapi.EventTypeNormal, "Steps",
+				fmt.Sprintf("Refs:%v Steps Error:%s", o.jobSpec.Refs, err))
 			return fmt.Errorf("could not run steps: %v", err)
 		}
 
 		for _, step := range postSteps {
 			if err := step.Run(ctx, o.dry); err != nil {
+				eventRecorder(client, o.namespace).Event(&coreapi.ObjectReference{Namespace: o.namespace}, coreapi.EventTypeNormal, "PostSteps",
```
smarterclayton: I would like the message to look more like the description we put on the namespace (see below)

droslean:
It seems that the record package needs those 2 dependencies. I added them manually, and I didn't update glide. 
It seems that glide can't get a single package and add it in the vendor's folder. There is an opened [issue](https://github.com/Masterminds/glide/issues/328) in glide repo for that.

cmd/ci-operator/main.go
droslean: @smarterclayton What output we should approach in case of multiple pulls or refs that have multiple pulls as well?

droslean: Something like 
``` go
eventRecorder(client, o.namespace).Event(&coreapi.ObjectReference{Namespace: o.namespace}, coreapi.EventTypeNormal, "References",
	fmt.Sprintf("Authors %s in Job: %s with baseSHA: %s\n", func() []string {
		var results []string
		for _, pull := range o.jobSpec.Refs.Pulls {
			out, err := json.Marshal(pull.Author + " with PR: " + strconv.Itoa(pull.Number))
			if err != nil {
				return []string{}
			}
			results = append(results, string(out))
		}
		return results
	}(), o.jobSpec.Job, o.jobSpec.Refs.BaseSHA))
```
With message output will look like:
```
Authors ["droslean with PR: 243" "clayton with PR: 221"] in Job: test-presubmit with baseSHA: 4b75c3617b0f591a8ca32c9ef311750d75b9a3d0
```

stevekuznetsov:
@smarterclayton please explain basic use-cases for event monitoring here, I can see some arguments for dumping structured JSON into the event body so we don't human-format things like pull number, author, etc ... but this depends on how we're using the events

smarterclayton:
So the minimal part would would be seeing an event stream on the cluster and see interspersed records of what succeed:

```
2018-08-24T17:30:29Z	CiJobStarted	Running job pull-ci-openshift-api-unit for PR https://github.com/openshift/origin/pull/534 in namespace ci-op-x3435 from author smarterclayton
2018-08-24T17:30:39Z	Scheduled	Successfully assigned ci-op-5xp76lcd/e2e-gcp to origin-ci-ig-n-4w66
2018-08-24T17:30:41Z	Created	Created container
2018-08-24T17:30:41Z	Pulled	Successfully pulled image "registry.svc.ci.openshift.org/ci-op-5xp76lcd/stable:tests"
2018-08-24T17:30:41Z	Pulling	pulling image "registry.svc.ci.openshift.org/ci-op-5xp76lcd/stable:tests"
2018-08-24T17:30:42Z	Pulling	pulling image "registry.svc.ci.openshift.org/ci-op-2018-08-24T17:31:30Z	Pulled	Successfully pulled image "busybox"
2018-08-24T17:31:30Z	Pulling	pulling image "busybox"
2018-08-24T17:31:30Z	Started	Started container
2018-08-24T17:32:10Z	CiJobFailed	Job pull-ci-openshift-api-unit failed for PR https://github.com/openshift/origin/pull/534 in namespace ci-op-x3435 from author smarterclayton
```

Because if I saw an event right above CiJobFailed like `NodeEvicted in namespace ci-op-x3435` I immediately know why.

cmd/ci-operator/main.go
```diff
@@ -422,17 +423,28 @@ func (o *options) Run() error {
...
+		runtimeObject := &coreapi.ObjectReference{Namespace: o.namespace}
+
+		eventRecorder.Event(runtimeObject, coreapi.EventTypeNormal, "CiJobStarted", eventJobDescription(o.jobSpec, o.namespace))
 		// execute the graph
 		suites, err := steps.Run(ctx, nodes, o.dry)
 		if err := o.writeJUnit(suites, "operator"); err != nil {
 			log.Printf("warning: Unable to write JUnit result: %v", err)
 		}
 		if err != nil {
+			eventRecorder.Event(runtimeObject, coreapi.EventTypeNormal, "CiJobFailed", eventJobDescription(o.jobSpec, o.namespace))
```
droslean: @smarterclayton That is not working because it returns before the Event will succeed. Is there any canonical way to wait for the event recorder to send the event?
I guess time.Sleep or add a WaitGroup is not a good idea.

stevekuznetsov:
Events are non-blocking so they are also best-effort. Not sure working around that somehow is a reasonable goal.

cmd/ci-operator/main.go
smarterclayton: I would recommend not trying to recalculate the message yourself, but share the description calculation string from below (`jobDescription`)

smarterclayton:
Yes, you should not try to change the async behavior of events.  As Steve noted, they are informational and best-effort.

cmd/ci-operator/main.go
droslean: @smarterclayton I created eventJobDescription() for this. The comment below are outdated.

droslean:
@stevekuznetsov @smarterclayton The problem is that if we won't do that, the Failed Event message will never occur. The program will exit before the event will complete.

smarterclayton:
You can do a sleep 1s at the very end before we exit main.

PR #102, opened 2018-08-22
Ensure target imagestream exists before promoting

stevekuznetsov:

/assign @smarterclayton

### DIFF
--- a/pkg/steps/promote.go
+++ b/pkg/steps/promote.go
@@ -102,9 +102,25 @@ func (s *promotionStep) Run(ctx context.Context, dry bool) error {
 		if valid == nil {
 			continue
 		}
+
+		name := fmt.Sprintf("%s%s", s.config.NamePrefix, dst)
+
+		_, err := s.dstClient.ImageStreams(s.config.Namespace).Get(name, meta.GetOptions{})
+		if errors.IsNotFound(err) {
+			_, err = s.dstClient.ImageStreams(s.config.Namespace).Create(&imageapi.ImageStream{
+				ObjectMeta: meta.ObjectMeta{
+					Name:      name,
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
@smarterclayton ping

pkg/steps/promote.go
```diff
@@ -102,9 +102,25 @@ func (s *promotionStep) Run(ctx context.Context, dry bool) error {
 		if valid == nil {
 			continue
 		}
+
+		name := fmt.Sprintf("%s%s", s.config.NamePrefix, dst)
+
+		_, err := s.dstClient.ImageStreams(s.config.Namespace).Get(name, meta.GetOptions{})
```
smarterclayton: You can just do a blind create and ignore already exists.  That would be simpler and easier to reason about.

smarterclayton: And ignore forbidden (because you might have update permission but not create).

stevekuznetsov: ACK

stevekuznetsov:
@smarterclayton actually no I thought about it and I disagree. If we see there is no imagestream, we should create one. If we cannot create one and we know one does not exist, we should fail. No job asking for `--promote` should succeed if the output image is not present, as is currently the case ...

smarterclayton:
Promotion is either going to succeed or fail.  The only permission ci-operator requires is the ability to update a tag.  If you don’t have create permissions and the stream doesn’t exist, the tag update will fail with the correct error.  If you do have permission to update and the stream exists, then you didn’t need create permission.

There’s no need to check the get before calling create.

smarterclayton:
Yeah, still don't agree with this change.

PR #103, opened 2018-08-22
When no release configuration is in place, have a shim prereq

smarterclayton:
images still create output tags, so [images] will continue work.
If no release tag configuration exists, use a simple step that just
creates stable.

@stevekuznetsov


stevekuznetsov:
Oh wait nvm we have chicken/egg issues
/hold cancel

stevekuznetsov:
So hilariously without a manual build there is no way to update the ci-operator image for this fix

PR #111, opened 2018-08-27
Add documentation on the use of template steps

stevekuznetsov:

/cc @petr-muller @bbguimaraes @coverprice @droslean

### DIFF
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -1,0 +1,103 @@
+# Using `Template`s as Execution Targets With CI Operator
+
+The CI Operator allows for users to configure tests running in a single container
+with the `tests` array. In many cases, however, running a command in a single
+container will not suffice. For more complicated tests, the CI Operator supports
+a `--template` option with which a test author may write black-boxed execution
+targets using OpenShift `Template`s. This document explains the interfaces expected
+for test `Template`s and best practices for how an author should configure their
+`Template`.
+
+The high-level contract for a `Template` execution target is that the `Template`
+will be processed into the test's `Namespace` once prerequisite `Build`s have
+occurred. Artifacts will be gathered from `artifacts` volumes on `Pod`s in
+the `Template`, if any are present. What work occurs in a `Template`'s `Pod`s is
+a black box as far as the CI Operator is concerned.
+
+## Parameters Available to `Template`s
+
+A number of parameters are available to the `Template` and will be provided if
+the `Template` is explicitly configured to consume them. The following parameters
+will be provided to `Template`s:
+
+ - `NAMESPACE`
+   The namespace in which your `Template` is processed. Target this namespace with
+   your objects.
+
+   | Example: | `ci-op-<input-hash>` |
+   | - | - |
+
+ - `IMAGE_FORMAT`
+   The image repository URL template using which your job's release images can
+   be pulled, formatted for the OpenShift Ansible `-e oreg_url` parameter.
+   `Template`s depending on this parameter will be processed only after all
+   images are built.
+
+   | Example: | `registry.svc.ci.openshift.org/ci-op-<input-hash>/stable:${component}` |
+   | - | - |
+
+
+ - `RPM_REPO`
+   The RPM repository URL from which your job's RPMs can be installed.
+   `Template`s depending on this parameter will be processed only after the
+   `rpm` image is built.
+
+   | Example: | `http://rpm-repo-ci-op-<input-hash>.svc.ci.openshift.org` |
+   | - | - |
+
+ - `IMAGE_<component>`
+   The image repository URL from which the `<component>` image can be pulled.
+   If a component name has hyphens, replace them with underscores. For instance,
+   the `$IMAGE_SOME_NAME` parameter would point to the registry URL for `some-name`.
+   Valid `<component>` images are any non-optional images defined in the `images`
+   array. Many parameters of this format can be provided for one `Template`.
+   `Template`s depending on these parameters will be processed only after the
+   `<component>` images are built.
+
+   | Example: | `registry.svc.ci.openshift.org/ci-op-<input-hash>/stable:<component>` |
+   | - | - |
+
+ - `LOCAL_IMAGE_<pipeline-tag>`
+   The image repository URL from which the `<pipeline-tag>` image can be pulled.
+   If a component name has hyphens, replace them with underscores. For instance,
+   the `$LOCAL_IMAGE_SOME_NAME` parameter would point to the registry URL for
+   `some-name`. Valid `<pipeline-tag>` images are `src`, `bin`, `test-bin`, or
+   `rpms`. Many parameters of this format can be provided for one `Template`.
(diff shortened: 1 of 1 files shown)


smarterclayton:
We could just check in the output generated by the help to a readme file as well.  This looks fine.

TEMPLATES.md
```diff
@@ -0,0 +1,101 @@
...
+is considered successful if every `Container` in every `Pod` in the `Template`
+definition starts at least once and terminates with a `0` exit code.
+
+### `Template` Artifact Output
+
+The CI Operator will add a `Container` named `artifacts` to any `Pod` defined in
+the `Template` that exposes a volume named `artifacts`. All files that are added
+to the volume will be uploaded to the appropriate GCS bucket for the job build
+that executed the `Template`. Failures to retrieve or upload artifacts will not
+impact the overall result of the job.
```
bbguimaraes: It's worth pointing out that this only happens when `--artifacts-dir` is set.

Interestingly, the first sentence is true even without the flag. I wonder why that is.

stevekuznetsov: /shrug

stevekuznetsov:
Want to link out to this doc, if there is more feedback we can address in a follow-up.

TEMPLATES.md
```diff
@@ -0,0 +1,103 @@
+# Using `Template`s as Execution Targets With CI Operator
```
coverprice: Nit: Top level directory docs tend to be the boilerplate stuff: README.md, LICENSE, etc. More detailed docs like this and the architecture docs should go in a Documentation directory. Then the filenames don't have to be so shouty :)

TEMPLATES.md
```diff
@@ -0,0 +1,103 @@
...
+
+The CI Operator allows for users to configure tests running in a single container
+with the `tests` array. In many cases, however, running a command in a single
+container will not suffice. For more complicated tests, the CI Operator supports
+a `--template` option with which a test author may write black-boxed execution
+targets using OpenShift `Template`s. This document explains the interfaces expected
```
coverprice: I'm gonna guess OpenShift Templates are a foreign concept to most readers, so perhaps link to the docs for them here.

TEMPLATES.md
```diff
@@ -0,0 +1,103 @@
...
+
+A number of parameters are available to the `Template` and will be provided if
+the `Template` is explicitly configured to consume them. The following parameters
+will be provided to `Template`s:
+
+ - `NAMESPACE`
+   The namespace in which your `Template` is processed. Target this namespace with
+   your objects.
+
+   | Example: | `ci-op-<input-hash>` |
```
coverprice: Because this whole thing is about Templates, it's really important for your examples to be clear when they refer to dynamic data. Right now it's not clear which parts are dynamic and which parts are literal.

Your examples are not (strictly speaking) examples; they're format definitions (I think). If I'm describing the output of `md5sum`, then I might say:

    Output format: <hash> <filename>
    Example: f3a0efce6fa8d4437c0a8392f4622db8 foobar

Your examples seem to combine the two, and mix and match `<thisformat>` and `${thisformat}`. This makes the `IMAGE_FORMAT` definition below confusing to read; it's not clear whether `${component}` is a literal or not.

To improve this, I'd change the "Examples" to actual full-blown examples; don't redact them in any way.

TEMPLATES.md
```diff
@@ -0,0 +1,103 @@
...
+
+The high-level contract for a `Template` execution target is that the `Template`
+will be processed into the test's `Namespace` once prerequisite `Build`s have
+occurred. Artifacts will be gathered from `artifacts` volumes on `Pod`s in
+the `Template`, if any are present. What work occurs in a `Template`'s `Pod`s is
+a black box as far as the CI Operator is concerned.
+
+## Parameters Available to `Template`s
+
+A number of parameters are available to the `Template` and will be provided if
```
coverprice: This sentence is confusing to me. Unless I'm wrong, isn't it just simpler to say "The following template variables are available:"? As it stands, this sentence seems to suggest that they won't be available unless the Template explicitly asks for them, which doesn't sound like any templating system I've ever heard of.

TEMPLATES.md
```diff
@@ -0,0 +1,103 @@
...
+ - `RPM_REPO`
+   The RPM repository URL from which your job's RPMs can be installed.
+   `Template`s depending on this parameter will be processed only after the
+   `rpm` image is built.
+
+   | Example: | `http://rpm-repo-ci-op-<input-hash>.svc.ci.openshift.org` |
+   | - | - |
+
+ - `IMAGE_<component>`
+   The image repository URL from which the `<component>` image can be pulled.
```
coverprice: There's mixing and matching of jargon here; "component" refers to "component image", right? I'd just use "image", and clarify where that comes from. E.g. "`IMAGE_<image>` expands to the image repository URL from which image `<image>` can be pulled. The list of available values for `<image>` is defined in the CI Operator config file under `images:`. Dashes in the config file image name are converted to underscores."

TEMPLATES.md
```diff
@@ -0,0 +1,103 @@
...
+   `some-name`. Valid `<pipeline-tag>` images are `src`, `bin`, `test-bin`, or
+   `rpms`. Many parameters of this format can be provided for one `Template`.
+   `Template`s depending on these parameters will be processed only after the
+   `<pipeline-tag>` images are built.
+
+   | Example: | `registry.svc.ci.openshift.org/ci-op-<input-hash>/pipeline:<pipeline-tag>` |
+   | - | - |
+
+ - `JOB_NAME`
+   The job name as provided by Prow in the `$JOB_SPEC`.
```
coverprice: Where is `$JOB_SPEC` defined?

TEMPLATES.md
```diff
@@ -0,0 +1,103 @@
...
+
+ - `JOB_NAME_SAFE`
+   The `$JOB_NAME` transformed to be a valid Kubernetes object identifier.
+
+ - `JOB_NAME_HASH`
+   A short hash of the `$JOB_NAME`. Append this to the `$NAMESPACE` to create an
+   identifier that is unique across all jobs for the same inputs. Note that this
+   identifier will not be unique across multiple builds of the same job for the
```
coverprice: "this identifier" refers to the `JOB_NAME_HASH` or the concatenation of `$NAMESPACE` and `$JOB_NAME_HASH`?

TEMPLATES.md
```diff
@@ -0,0 +1,103 @@
...
+   same inputs.
+
+## Expected Output From `Template`s
+
+The CI Operator attempts to determine the result of the `Template`'s execution
```
coverprice: This sentence doesn't really communicate anything... I'd change the heading to "Output" and remove this sentence.

TEMPLATES.md
```diff
@@ -0,0 +1,103 @@
...
+as well as assembling output artifacts.
+
+### `Template` Execution Outcome
+
```
coverprice: "CI Operator determines the success of the `Template`'s execution by examining Pod exit codes."

PR #117, opened 2018-08-30
Correctly return error when creating build

stevekuznetsov:

/assign @smarterclayton


smarterclayton:
/lgtm

we should run errcheck

PR #120, opened 2018-08-31
Add custom base image option.

droslean:
Implements the ability to use an image built from a Dockerfile located in a git source.

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -1,18 +1,27 @@
 package api
 
-import (
-	"fmt"
-)
+import "errors"
 
 // Validate validates config
 func (config *ReleaseBuildConfiguration) Validate() error {
+	buildRootImage := config.InputConfiguration.BuildRootImage
+	testBaseImage := config.InputConfiguration.TestBaseImage
 	if config.Tests != nil {
 		for _, test := range config.Tests {
 			if test.As == "images" {
-				return fmt.Errorf("Test should not be called 'images' because it gets confused with '[images]' target")
+				return errors.New("test should not be called 'images' because it gets confused with '[images]' target")
 			}
 		}
+	} else if buildRootImage != nil && testBaseImage != nil {
+		return errors.New("both build_root and test_base_image cannot be set")
+	} else if buildRootImage != nil && testBaseImage == nil {
+		if buildRootImage.ProjectImageBuild != nil && buildRootImage.ImageStreamTagReference != nil {
+			return errors.New("both git_source_image and image_stream_tag cannot be set for the build_root")
+		} else if buildRootImage.ProjectImageBuild == nil && buildRootImage.ImageStreamTagReference == nil {
+			return errors.New("you have to specify either git_source_image or image_stream_tag for the build_root")
+		}
+	} else if buildRootImage == nil && testBaseImage == nil {
+		return errors.New("no build_root or test_base_image has been set")
 	}
-
 	return nil
 }
--- a/pkg/api/config_test.go
+++ b/pkg/api/config_test.go
@@ -32,7 +32,57 @@ func TestValidate(t *testing.T) {
 				},
 			},
 			expectedValid: false,
-			expectedError: "Test should not be called 'images' because it gets confused with '[images]' target",
+			expectedError: "test should not be called 'images' because it gets confused with '[images]' target",
+		},
+		{
+			id: "both git_source_image and image_stream_tag in build_root defined causes error",
+			config: ReleaseBuildConfiguration{
+				InputConfiguration: InputConfiguration{
+					BuildRootImage: &BuildRootImageConfiguration{
+						ImageStreamTagReference: &ImageStreamTagReference{
+							Cluster:   "test_cluster",
+							Namespace: "test_namespace",
+							Name:      "test_name",
+							Tag:       "test",
+						},
+						ProjectImageBuild: &ProjectDirectoryImageBuildInputs{
+							ContextDir:     "/",
+							DockerfilePath: "Dockerfile.test",
+						},
+					},
+				},
+			},
+			expectedValid: false,
+			expectedError: "both git_source_image and image_stream_tag cannot be set for the build_root",
+		},
+		{
+			id: "build root without any content causes an error",
+			config: ReleaseBuildConfiguration{
+				InputConfiguration: InputConfiguration{
+					BuildRootImage: &BuildRootImageConfiguration{},
+				},
+			},
+			expectedValid: false,
+			expectedError: "you have to specify either git_source_image or image_stream_tag for the build_root",
+		},
+		{
+			id: "build_root and test_base_image defined causes an error",
+			config: ReleaseBuildConfiguration{
+				InputConfiguration: InputConfiguration{
+					BuildRootImage: &BuildRootImageConfiguration{},
+					TestBaseImage:  &ImageStreamTagReference{},
+				},
+			},
+			expectedValid: false,
+			expectedError: "both build_root and test_base_image cannot be set",
+		},
+		{
+			id: "build_root and test_base_image not defined causes an error",
+			config: ReleaseBuildConfiguration{
+				InputConfiguration: InputConfiguration{},
+			},
+			expectedValid: false,
+			expectedError: "no build_root or test_base_image has been set",
 		},
 	}
 
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -123,9 +123,21 @@ type InputConfiguration struct {
 	// build-time dependencies for the project.
 	TestBaseImage *ImageStreamTagReference `json:"test_base_image,omitempty"`
 
+	// BuildRootImage supports two ways to get the image that
+	// the pipeline will caches on. The one way is to take the reference
+	// from an image stream, and the other from a dockerfile.
+	BuildRootImage *BuildRootImageConfiguration `json:"build_root,omitempty"`
+
 	// ReleaseTagConfiguration determines how the
 	// full release is assembled.
 	ReleaseTagConfiguration *ReleaseTagConfiguration `json:"tag_specification,omitempty"`
+}
+
+// BuildRootImageConfiguration holds the two ways of using a base image
+// that the pipeline will caches on.
+type BuildRootImageConfiguration struct {
+	ImageStreamTagReference *ImageStreamTagReference          `json:"image_stream_tag,omitempty"`
+	ProjectImageBuild       *ProjectDirectoryImageBuildInputs `json:"project_image_build,omitempty"`
 }
 
 // ImageStreamTagReference identifies an ImageStreamTag
@@ -224,6 +236,7 @@ type StepConfiguration struct {
 	OutputImageTagStepConfiguration             *OutputImageTagStepConfiguration             `json:"output_image_tag_step,omitempty"`
 	ReleaseImagesTagStepConfiguration           *ReleaseTagConfiguration                     `json:"release_images_tag_step,omitempty"`
 	TestStepConfiguration                       *TestStepConfiguration                       `json:"test_step,omitempty"`
+	ProjectDirectoryImageBuildInputs            *ProjectDirectoryImageBuildInputs            `json:"project_directory_image_build_inputs,omitempty"`
 }
 
 // InputImageTagStepConfiguration describes a step that
@@ -318,6 +331,16 @@ type ProjectDirectoryImageBuildStepConfiguration struct {
 	From PipelineImageStreamTagReference `json:"from"`
 	To   PipelineImageStreamTagReference `json:"to"`
 
+	ProjectDirectoryImageBuildInputs `json:",inline"`
+
+	// Optional means the build step is not built, published, or
+	// promoted unless explicitly targeted. Use for builds which
+	// are invoked only when testing certain parts of the repo.
+	Optional bool `json:"optional"`
+}
+
+// ProjectDirectoryImageBuildInputs holds inputs for an image build from the repo under test
+type ProjectDirectoryImageBuildInputs struct {
 	// ContextDir is the directory in the project
 	// from which this build should be run.
 	ContextDir string `json:"context_dir"`
@@ -330,11 +353,6 @@ type ProjectDirectoryImageBuildStepConfiguration struct {
 	// that will populate the build context for the Dockerfile or
 	// alter the input image for a multi-stage build.
 	Inputs map[string]ImageBuildInputs `json:"inputs"`
-
-	// Optional means the build step is not built, published, or
-	// promoted unless explicitly targeted. Use for builds which
-	// are invoked only when testing certain parts of the repo.
-	Optional bool `json:"optional"`
 }
 
 // ImageBuildInputs is a subset of the v1 OpenShift Build API object
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -118,6 +118,8 @@ func FromConfig(
 			step = steps.SourceStep(*rawStep.SourceStepConfiguration, config.Resources, buildClient, srcClient, imageClient, jobSpec)
 		} else if rawStep.ProjectDirectoryImageBuildStepConfiguration != nil {
 			step = steps.ProjectDirectoryImageBuildStep(*rawStep.ProjectDirectoryImageBuildStepConfiguration, config.Resources, buildClient, imageClient, jobSpec)
+		} else if rawStep.ProjectDirectoryImageBuildInputs != nil {
+			step = steps.GitSourceStep(*rawStep.ProjectDirectoryImageBuildInputs, config.Resources, buildClient, imageClient, jobSpec)
 		} else if rawStep.RPMImageInjectionStepConfiguration != nil {
 			step = steps.RPMImageInjectionStep(*rawStep.RPMImageInjectionStepConfiguration, config.Resources, buildClient, routeGetter, imageClient, jobSpec)
 		} else if rawStep.RPMServeStepConfiguration != nil {
@@ -259,6 +261,18 @@ func stepConfigsForBuild(config *api.ReleaseBuildConfiguration, jobSpec *api.Job
 		config.InputConfiguration.BaseRPMImages[alias] = target
 	}
 
+	if target := config.InputConfiguration.BuildRootImage; target != nil {
+		if isTagRef := target.ImageStreamTagReference; isTagRef != nil {
+			buildSteps = append(buildSteps, createStepConfigForTagRefImage(*isTagRef, jobSpec))
+		} else if gitSourceRef := target.ProjectImageBuild; gitSourceRef != nil {
+			buildSteps = append(buildSteps, createStepConfigForGitSource(*gitSourceRef, jobSpec))
+		}
+	}
+
+	if target := config.InputConfiguration.TestBaseImage; target != nil {
+		buildSteps = append(buildSteps, createStepConfigForTagRefImage(*target, jobSpec))
+	}
+
 	buildSteps = append(buildSteps, api.StepConfiguration{SourceStepConfiguration: &api.SourceStepConfiguration{
 		From:      api.PipelineImageStreamTagReferenceRoot,
 		To:        api.PipelineImageStreamTagReferenceSource,
@@ -271,24 +285,6 @@ func stepConfigsForBuild(config *api.ReleaseBuildConfiguration, jobSpec *api.Job
 		},
 		ClonerefsPath: "/clonerefs",
 	}})
-
-	if target := config.InputConfiguration.TestBaseImage; target != nil {
-		if target.Namespace == "" {
-			target.Namespace = jobSpec.BaseNamespace
-		}
-		if target.Name == "" {
-			target.Name = fmt.Sprintf("%s-test-base", jobSpec.Refs.Repo)
-		}
-		alias := target.As
-		if len(alias) == 0 {
-			alias = target.Tag
-		}
-
-		buildSteps = append(buildSteps, api.StepConfiguration{InputImageTagStepConfiguration: &api.InputImageTagStepConfiguration{
-			BaseImage: *target,
-			To:        api.PipelineImageStreamTagReferenceRoot,
-		}})
-	}
 
 	if len(config.BinaryBuildCommands) > 0 {
 		buildSteps = append(buildSteps, api.StepConfiguration{PipelineImageCacheStepConfiguration: &api.PipelineImageCacheStepConfiguration{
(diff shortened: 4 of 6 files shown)


pkg/api/types.go
```diff
@@ -123,11 +123,21 @@ type InputConfiguration struct {
...
+
 	// ReleaseTagConfiguration determines how the
 	// full release is assembled.
 	ReleaseTagConfiguration *ReleaseTagConfiguration `json:"tag_specification,omitempty"`
 }
 
+// GitSourceReference
+type GitSourceReference struct {
+	ContextDir     string `json:"contextDir,omitempty"`
+	Uri            string `json:"uri,omitempty"`
```
stevekuznetsov: The URI and Ref are provided from Prow in the job spec and do not need to be exposed here

pkg/api/types.go
```diff
@@ -123,11 +123,21 @@ type InputConfiguration struct {
 	// build-time dependencies for the project.
 	TestBaseImage *ImageStreamTagReference `json:"test_base_image,omitempty"`
 
+	CustomTestBaseImage *GitSourceReference `json:"custom_test_base_image,omitempty"`
```
stevekuznetsov: As noted on slack this should be

```go
type InputConfiguration struct {
    // ...
    BuildRootImage *BuildRootImageConfiguration `json:"build_root,omitempty"`
    // ...
}

type BuildRootImageConfiguration struct {
    ExistingImageReference *ImageStreamTagReference `json:"existing_image,omitempty"`
    ProjectImageConfiguration *ProjectDirectoryImageBuildStepConfiguration `json:"project_image,omitempty"`
}
```

pkg/steps/git_source.go
```diff
@@ -0,0 +1,143 @@
...
+	return nil
+	//return []api.StepLink{api.InternalImageLink(s.gitConfig.BaseImage)}
+}
+func (s *gitSourceStep) Creates() []api.StepLink {
+	return []api.StepLink{api.InternalImageLink(s.gitConfig.To)}
+}
+
+func (s *gitSourceStep) Provides() (api.ParameterMap, api.StepLink) {
+	return api.ParameterMap{
+		"LOCAL_IMAGE_SRC": func() (string, error) {
```
stevekuznetsov: This is wrong, it provides base

pkg/api/types.go
```diff
@@ -123,11 +123,32 @@ type InputConfiguration struct {
...
+	ImageStreamTagReference *ImageStreamTagReference `json:"image_stream_tag,omitempty"`
+	GitSourceReference      *GitSourceReference      `json:"git_source_image,omitempty"`
+}
+
+// GitSourceReference identifies the location of a Dockerfile that will
+// be used as source for a build.
+type GitSourceReference struct {
+	ContextDir string `json:"context_dir,omitempty"`
+	Uri        string `json:"uri,omitempty"`
+	Ref        string `json:"ref,omitempty"`
```
bbguimaraes: Do we want to support files outside the repository being tested?

pkg/steps/git_source.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (s *gitSourceStep) Requires() []api.StepLink { return nil }
+
+func (s *gitSourceStep) Creates() []api.StepLink {
+	return []api.StepLink{api.InternalImageLink(s.gitConfig.To)}
+}
+
+func (s *gitSourceStep) Provides() (api.ParameterMap, api.StepLink) {
+	return api.ParameterMap{
+		"BUILD_ROOT_IMAGE": func() (string, error) {
```
bbguimaraes: Still not correct, this should be the name of the parameter expected by a template step.

pkg/steps/git_source.go
```diff
@@ -0,0 +1,140 @@
...
+				Output: buildapi.BuildOutput{
+					To: &coreapi.ObjectReference{
+						Kind:      "ImageStreamTag",
+						Namespace: jobSpec.Namespace(),
+						Name:      fmt.Sprintf("%s:%s", PipelineImageStream, toTag),
+					},
+				},
+			},
+		},
+	}
```
bbguimaraes: Only the `BuildSource` is different from the existing source build, copying the code creates a lot of unnecessary duplication.

pkg/api/types.go
droslean: In case of those are empty, I get them from the current org/repo that it's referred to job_spec. I left those values there for review. What if the Dockerfiles are in a different repo?

@stevekuznetsov has already mentioned that we don't have to support files outside the repo.

petr-muller:
I know it's WIP but a lack of at least very brief description what is the intent of the change makes this a bit harder to review.

pkg/api/types.go
stevekuznetsov: Comment on `ProjectDirectoryImageBuildStepConfiguration` is outstanding

stevekuznetsov:
One comment https://github.com/openshift/ci-operator/pull/120#discussion_r214709603

pkg/api/types.go
```diff
@@ -123,11 +123,32 @@ type InputConfiguration struct {
 	// build-time dependencies for the project.
 	TestBaseImage *ImageStreamTagReference `json:"test_base_image,omitempty"`
 
+	// BuildRootImage supports two ways to get the image that
+	// the pipeline will caches on. The one way is to take the reference
+	// from an image stream, and the other from a dockerfile.
+	BuildRootImage *BuildRootImageConfiguration `json:"build_root,omitempty"`
```
smarterclayton: Why is this special?  Why not just key off an image called "root"?

pkg/api/types.go
```diff
@@ -123,11 +123,32 @@ type InputConfiguration struct {
...
+
 	// ReleaseTagConfiguration determines how the
 	// full release is assembled.
 	ReleaseTagConfiguration *ReleaseTagConfiguration `json:"tag_specification,omitempty"`
 }
 
+// BuildRootImageConfiguration holds the two ways of using a base image
+// that the pipeline will caches on.
+type BuildRootImageConfiguration struct {
```
smarterclayton: What is the exact use case this is solving?  Why is this is any different than any other "image"?

smarterclayton:
/hold

This breaks input caching, and we'd have to lookup the source location and include that in the job spec cache.  I'd prefer this just be implemented by putting your extra source in a separate repo if you need different git source.

pkg/api/types.go
droslean: Trying to give to user two options in how he will pass the base image. And the config can be like this:
```
{
  "build_root": {
    "git_source_image":{
      "dockerfile": "Dockerfile"
    }, 
or 
    "image_stream_tag":{
      "cluster": "https://api.ci.openshift.org",
      "namespace": "openshift",
      "name": "release",
      "tag": "golang-1.10"
    }
  }
}
```

If the Uri is not specified as the example above, it will get it from $JOB_SPEC

pkg/steps/git_source.go
droslean: Also, there is a slight difference in the strategy of the build. I split them up because I didn't want to construct 2 different things with the same method, and add checks and change the source and strategy depending on which in nil or not.

smarterclayton:
Yeah, rename was fine to me.  Also, as we increasingly use multistage
builds the base image will be less important - it’s only a point in time
where we have this.


On Sep 4, 2018, at 6:24 AM, Bruno Barcarol Guimarães <
notifications@github.com> wrote:

@smarterclayton <https://github.com/smarterclayton>: this field is a rename
of test_base_image (because it causes confusion) while simultaneously
adding the possibility to build the image using a Dockerfile in the
repository instead of using an existing image stream. Corresponding JIRA
issue: https://jira.coreos.com/browse/DPTP-95.

It can't be handled the same as the other images because it's the root
image.

And I agree that breaking caching is a big reason not to support external
repositories.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/ci-operator/pull/120#issuecomment-418317366>,
or mute the thread

pkg/api/types.go
```diff
@@ -123,11 +123,30 @@ type InputConfiguration struct {
...
+	ImageStreamTagReference *ImageStreamTagReference `json:"image_stream_tag,omitempty"`
+	GitSourceReference      *GitSourceReference      `json:"git_source_image,omitempty"`
+}
+
+// GitSourceReference identifies the location of a Dockerfile that will
+// be used as source for a build.
+type GitSourceReference struct {
```
stevekuznetsov: This is now even closer to `ProjectDirectoryImageBuildStepConfiguration` -- let's factor out the common fields (`ContextDir`, `DockerfilePath` and `Inputs`) and use struct embedding to share the sub-struct between this use-case and the `ProjectDirectoryImageBuildStepConfiguration`.

pkg/defaults/defaults.go
```diff
@@ -256,19 +258,38 @@ func stepConfigsForBuild(config *api.ReleaseBuildConfiguration, jobSpec *api.Job
...
+			buildSteps = append(buildSteps, api.StepConfiguration{
+				InputGitSourceStepConfiguration: &api.InputGitSourceStepConfiguration{
+					BaseImage: *gitSourceRef,
+					To:        api.PipelineImageStreamTagReferenceRoot},
+			})
+		}
+	}
+
+	// This will be removed in future versions.
 	if target := config.InputConfiguration.TestBaseImage; target != nil {
```
stevekuznetsov: Validate that one or the other is provided but not both

pkg/steps/source.go
```diff
@@ -53,6 +53,7 @@ RUN git submodule update --init
 
 type sourceStep struct {
 	config             api.SourceStepConfiguration
+	gitConfig          *api.InputGitSourceStepConfiguration
```
stevekuznetsov: The `sourceStep` is what creates the `src` tag in the `pipeline` imagestream. The test base image is _not_ the same.

pkg/api/config.go
```diff
@@ -6,13 +6,22 @@ import (
...
 func (config *ReleaseBuildConfiguration) Validate() error {
+	buildRootImage := config.InputConfiguration.BuildRootImage
+	testBaseImage := config.InputConfiguration.TestBaseImage
 	if config.Tests != nil {
 		for _, test := range config.Tests {
 			if test.As == "images" {
 				return fmt.Errorf("Test should not be called 'images' because it gets confused with '[images]' target")
 			}
 		}
+	} else if buildRootImage != nil && testBaseImage == nil {
```
stevekuznetsov: This shouldn't be an `else`, just have it top-level

pkg/api/config.go
```diff
@@ -6,13 +6,22 @@ import (
...
+		if buildRootImage.GitSourceReference != nil && buildRootImage.ImageStreamTagReference != nil {
+			return fmt.Errorf("You can't use both git_source_image and image_stream_tag references")
```
stevekuznetsov: If you have no formatting directives, use `errors.New()`, not `fmt.Errorf()`

stevekuznetsov: Add some more context -- `"both git_source_image and `image_stream_tag` cannot be set for the built_root"`

pkg/api/config.go
```diff
@@ -6,13 +6,22 @@ import (
...
+		} else if buildRootImage.GitSourceReference == nil && buildRootImage.ImageStreamTagReference == nil {
+			return fmt.Errorf("You have to specify either git_source_image or image_stream_tag references")
+		}
+	} else if buildRootImage == nil && testBaseImage == nil {
```
stevekuznetsov: What happens if someone passes in both the build root and test base image?

pkg/api/types.go
```diff
@@ -335,6 +347,9 @@ type ProjectDirectoryImageBuildStepConfiguration struct {
 	// promoted unless explicitly targeted. Use for builds which
 	// are invoked only when testing certain parts of the repo.
 	Optional bool `json:"optional"`
+
+	// IsGitSource defines if the build will be created with a git source.
+	IsGitSource bool `json:"is_git_source"`
```
stevekuznetsov: I think the struct embedding would be more appropriate -- factor out the common bits and use them as they need to be used. An API that provides one shape with a switch for behavior is confusing.

pkg/defaults/defaults.go
```diff
@@ -256,19 +260,40 @@ func stepConfigsForBuild(config *api.ReleaseBuildConfiguration, jobSpec *api.Job
...
-			Cluster:   "https://api.ci.openshift.org",
-			Namespace: "ci",
-			Name:      "clonerefs",
-			Tag:       "latest",
-		},
-		ClonerefsPath: "/clonerefs",
-	}})
+	if target := config.InputConfiguration.BuildRootImage; target != nil {
+		if isTagRef := target.ImageStreamTagReference; isTagRef != nil {
+			if isTagRef.Namespace == "" {
```
stevekuznetsov: Factor this code out to share it

pkg/api/types.go
droslean: ProjectDirectoryImageBuildStepConfiguration holds already the fields that we need.

droslean: it is being used here
https://github.com/openshift/ci-operator/pull/120/files#diff-dd3690bad12236edb7359fb0e2df4ccdR119

pkg/api/config_test.go
```diff
@@ -32,7 +32,38 @@ func TestValidate(t *testing.T) {
...
+							Dockerfile: "Dockerfile.test",
+						},
+					},
+				},
+			},
+			expectedValid: false,
+			expectedError: "both git_source_image and image_stream_tag cannot be set for the build_root",
+		},
+		{
+			id: "EmptyBuildRoot",
```
stevekuznetsov: test names should try to convey what and why -- so:

```go
id: "build root without any content causes an error"
```

This message needs to make sense for someone who has not written the test

pkg/api/config_test.go
```diff
@@ -32,7 +32,38 @@ func TestValidate(t *testing.T) {
 				},
 			},
 			expectedValid: false,
-			expectedError: "Test should not be called 'images' because it gets confused with '[images]' target",
+			expectedError: "test should not be called 'images' because it gets confused with '[images]' target",
+		},
+		{
+			id: "BuildRoot",
```
stevekuznetsov: You introduce four unique error conditions -- test for all of them, as well as the correct case

pkg/api/types.go
```diff
@@ -335,6 +354,8 @@ type ProjectDirectoryImageBuildStepConfiguration struct {
 	// promoted unless explicitly targeted. Use for builds which
 	// are invoked only when testing certain parts of the repo.
 	Optional bool `json:"optional"`
+
+	*GitSourceReference `json:",inline"`
```
stevekuznetsov: By adding this and not replacing the current fields you are duplicating the API -- we should have this _instead_ of the current fields in this struct.

pkg/defaults/defaults.go
```diff
@@ -256,37 +260,62 @@ func stepConfigsForBuild(config *api.ReleaseBuildConfiguration, jobSpec *api.Job
...
```
stevekuznetsov: Everything that takes the `isTagRef` and creates the `InputImageTagStepConfiguration` is duplicated here and for the `TestBaseImage` -- factor this to share logic.

pkg/steps/git_source.go
```diff
@@ -0,0 +1,65 @@
...
+func (s *gitSourceStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, error) {
+	return s.jobSpec.Inputs(), nil
+}
+
+func (s *gitSourceStep) Run(ctx context.Context, dry bool) error {
+	return handleBuild(s.buildClient, buildFromSource(s.jobSpec, "", s.config.To, buildapi.BuildSource{
+		Type:       buildapi.BuildSourceGit,
+		ContextDir: s.config.ContextDir,
+		Git: &buildapi.GitBuildSource{
+			URI: fmt.Sprintf("https://github.com/%s/%s.git", s.jobSpec.Refs.Org, s.jobSpec.Refs.Repo),
```
stevekuznetsov: How will we handle private repositories in the future?

pkg/steps/git_source.go
```diff
@@ -0,0 +1,65 @@
...
+	buildClient BuildClient
+	jobSpec     *api.JobSpec
+}
+
+func (s *gitSourceStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, error) {
+	return s.jobSpec.Inputs(), nil
+}
+
+func (s *gitSourceStep) Run(ctx context.Context, dry bool) error {
+	return handleBuild(s.buildClient, buildFromSource(s.jobSpec, "", s.config.To, buildapi.BuildSource{
```
stevekuznetsov: That empty string looks a tiny bit suspicious -- what is it?

pkg/steps/git_source.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (s *gitSourceStep) Requires() []api.StepLink { return nil }
+
+func (s *gitSourceStep) Creates() []api.StepLink {
+	return []api.StepLink{api.InternalImageLink(s.config.To)}
+}
+
+func (s *gitSourceStep) Provides() (api.ParameterMap, api.StepLink) {
+	return nil, nil
```
stevekuznetsov: You need to fill these in

pkg/steps/git_source.go
droslean: its a From tag. In this case, I am not using it.

pkg/steps/git_source.go
droslean: add Uri and Ref fields in the test config?

pkg/api/types.go
droslean: So this should be removed, as it was in the previous commits?

pkg/steps/git_source.go
droslean: Do we need this for the base image? We are not providing anything already.

pkg/api/types.go
stevekuznetsov: This is the last outstanding comment. Right now you have the `ContextDir` and the `DockerfilePath` in this configuration twice. You should remove the `ProjectDirectoryImageBuildStepConfiguration.ContextDir` and use only `ProjectDirectoryImageBuildStepConfiguration.GitSourceReference.ContextDir`. With the `json:",inline"` tag, the same JSON (or YAML) will get serialized correctly, since the embedded field uses the same config path.

stevekuznetsov:
```
@droslean I will try to lay out the reasoning behind struct embedding and how I would have expected to see it executed here.

The flow of thought is something like this:
 - we would like to expose an API for users to build their base image from their repo
 - we already expose an API for building images from repos
 - the current API is a strict superset of the prior API
 - instead of duplicating API fields and descriptions, we should share them between the two APIs

So, concretely, I would expect the following structs:

```go
type ReleaseBuildConfiguration struct {
	// ...
	Images []ProjectDirectoryImageBuildStepConfiguration `json:"images,omitempty"`
}

type InputConfiguration struct {
	// ...
	BuildRootImage *BuildRootImageConfiguration `json:"build_root,omitempty"`
}

// BuildRootImageConfiguration configures the base image for the pipeline cache.
type BuildRootImageConfiguration struct {
	ImageStreamTagReference *ImageStreamTagReference          `json:"image_stream_tag,omitempty"`
	ProjectImageBuild       *ProjectDirectoryImageBuildInputs `json:"project_image_build,omitempty"`
}

// ProjectDirectoryImageBuildStepConfiguration describes an
// image build from a directory in a component project.
type ProjectDirectoryImageBuildStepConfiguration struct {
	From PipelineImageStreamTagReference `json:"from"`
	To   PipelineImageStreamTagReference `json:"to"`

	ProjectDirectoryImageBuildInputs `json:",inline"`

	// Optional means the build step is not built, published, or
	// promoted unless explicitly targeted. Use for builds which
	// are invoked only when testing certain parts of the repo.
	Optional bool `json:"optional"`
}

// ProjectDirectoryImageBuildInputs holds inputs for an image build from the repo under test
type ProjectDirectoryImageBuildInputs struct {
	// ContextDir is the directory in the project
	// from which this build should be run.
	ContextDir string `json:"context_dir"`

	// DockerfilePath is the path to a Dockerfile in the
	// project to run relative to the context_dir.
	DockerfilePath string `json:"dockerfile_path"`

	// Inputs is a map of tag reference name to image input changes
	// that will populate the build context for the Dockerfile or
	// alter the input image for a multi-stage build.
	Inputs map[string]ImageBuildInputs `json:"inputs"`	
}
```

The new `ProjectDirectoryImageBuildInputs` structure contains the common input specification between the two similar steps.

Critically, though, is is _not valid_ for a user to provide simply those inputs for an image built from their repo using the `images` list. We continue to expose the exact same API surface for those images. We simply share that data with the root image build.

Then, when we are parsing the configuration the user has provided (in `pkg/defaults/defaults.go#stepConfigsForBuild()`) we know that every single configuration that a user has presented in the `images` list _will_ be translated to a step with a `ProjectDirectoryImageBuildStepConfiguration`. When looping through the list of raw steps, we similarly know that if a user has provided a `ProjectDirectoryImageBuildStepConfiguration`, they are asking for an image build. We have not changed the user-facing API for these actions in any way, we have simply shuffled the structure of the data on the back-end so that common concepts are shared when they are identical. We are _not_ exposing a new API for building images from the repo under test generically, so the only time a user will ever provide us with a simple `ProjectDirectoryImageBuildInputs` is in the context of the root image build, where they are not allowed to configure things like `from` and `to`. In that case and only in that case do we run the build using the git source. It's _not_ valid to expose the entire `ProjectDirectoryImageBuildStepConfiguration` in the root image build, as it is invalid for that root image build to output to any `ImageStreamTag` other than `pipeline:root`, it is invalid for there to be a configurable `from` for the root, and it is invalid for the root to be optional. 
```

File: pkg/defaults/defaults.go [hunk 24]
```diff
@@ -399,3 +397,33 @@ func stepConfigsForBuild(config *api.ReleaseBuildConfiguration, jobSpec *api.Job
...
+
+	return api.StepConfiguration{
+		InputImageTagStepConfiguration: &api.InputImageTagStepConfiguration{
+			BaseImage: target,
+			To:        api.PipelineImageStreamTagReferenceRoot,
+		}}
+}
+
+func createStepConfigForGitSource(target api.ProjectDirectoryImageBuildInputs, jobSpec *api.JobSpec) api.StepConfiguration {
+	if target.ContextDir == "" {
```
stevekuznetsov: If this is not provided by the user, I don't think it's valid to default it

pkg/steps/git_source.go
```diff
@@ -0,0 +1,65 @@
...
+	return s.jobSpec.Inputs(), nil
+}
+
+func (s *gitSourceStep) Run(ctx context.Context, dry bool) error {
+	return handleBuild(s.buildClient, buildFromSource(s.jobSpec, "", api.PipelineImageStreamTagReferenceRoot, buildapi.BuildSource{
+		Type:       buildapi.BuildSourceGit,
+		ContextDir: s.config.ContextDir,
+		Git: &buildapi.GitBuildSource{
+			URI: fmt.Sprintf("https://github.com/%s/%s.git", s.jobSpec.Refs.Org, s.jobSpec.Refs.Repo),
+			Ref: s.jobSpec.Refs.BaseRef,
```
stevekuznetsov: Technically this should be the merge commit of all the pulls and the base ref, but we have no way to do that with the OpenShift `Build`. We also can't just use the PR commit as that won't work on batches. We should add to the doc that you will get the build from the latest of the target branch, which is the same behavior as with a separate config. So technically still annoying but not a regression. Still possible to change the image build in the repo in a way where it won't break tests on the PR but will cause all subsequent tests to fail.

stevekuznetsov:
Perfect
/hold cancel
/lgtm

stevekuznetsov:
@origin-ci-robot that is one of my favorite instagram accounts!@

PR #125, opened 2018-09-04
Add test types

bbguimaraes:
Required for openshift/ci-operator-prowgen#4.

/hold

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -274,6 +274,7 @@ type PipelineImageCacheStepConfiguration struct {
 type TestStepConfiguration struct {
 	// As is the name of the test.
 	As string `json:"as"`
+	// TODO remove when the migration is completed
 	// From is the image stream tag in the pipeline to run this
 	// command in.
 	From PipelineImageStreamTagReference `json:"from"`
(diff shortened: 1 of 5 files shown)


pkg/api/types.go
```diff
@@ -276,6 +277,65 @@ type TestStepConfiguration struct {
...
+// Types of clusters supported by the provisioners.
+const (
+	ClusterTypeAWS ClusterType = "aws"
+	ClusterTypeGCP             = "gcp"
+)
+
+// ClusterTestConfiguration describes a test that provisions
+// a cluster and runs a command in it.
+type ClusterTestConfiguration struct {
+	ClusterType ClusterType `json:"cluster_type"`
```
stevekuznetsov: @smarterclayton convince me that end users don't want this to be `target_cloud`

pkg/api/types.go
```diff
@@ -276,6 +277,65 @@ type TestStepConfiguration struct {
...
+}
+
+// OpenshiftAnsibleClusterTestConfiguration describes a test
+// that provisions a cluster using openshift-ansible and runs
+// conformance tests.
+type OpenshiftAnsibleClusterTestConfiguration struct {
```
stevekuznetsov: Do we not want the upgrade scenarios in here, too?

pkg/defaults/defaults.go
```diff
@@ -387,8 +387,12 @@ func stepConfigsForBuild(config *api.ReleaseBuildConfiguration, jobSpec *api.Job
...
 	}
 
-	for i := range config.Tests {
-		buildSteps = append(buildSteps, api.StepConfiguration{TestStepConfiguration: &config.Tests[i]})
+	for _, test := range config.Tests {
+		// TODO remove when the migration is completed
+		if len(test.From) == 0 {
+			continue
+		}
+		buildSteps = append(buildSteps, api.StepConfiguration{TestStepConfiguration: &test})
```
stevekuznetsov: I am pretty sure taking the pointer to this `test` var won't work since it's reassigned in every loop iteration, which is why we were using `i` last time.

pkg/defaults/defaults.go
```diff
@@ -387,8 +387,12 @@ func stepConfigsForBuild(config *api.ReleaseBuildConfiguration, jobSpec *api.Job
 		}
 	}
 
-	for i := range config.Tests {
-		buildSteps = append(buildSteps, api.StepConfiguration{TestStepConfiguration: &config.Tests[i]})
+	for _, test := range config.Tests {
+		// TODO remove when the migration is completed
```
stevekuznetsov: Instead of ignoring it, you should set the stage for someone to use the new format. So

```go
if config[i].ContainerTestConfiguration != nil  && len(config[i].From) == 0 {
    config[i].From = config[i].ContainerTestConfiguration.From
}
```

Also add validation to ensure that `config[i].ContainerTestConfiguration == nil` XOR `len(config[i].From) == 0`

pkg/api/types.go
bbguimaraes: I think I mentioned this earlier today: I left the upgrade and all the azure templates out for now because they change every time I look. We can always add them later.

pkg/defaults/defaults.go
bbguimaraes: Yeah, I edited too quickly, that surely won't work as it is.

pkg/defaults/defaults.go
bbguimaraes: My thought was to do it in a separate PR, but it doesn't really matter. I'll add it here.

pkg/api/types.go
stevekuznetsov: ACK

pkg/defaults/defaults.go
stevekuznetsov: No preference on my part -- let's do it here

pkg/api/types.go
bbguimaraes: ; )

pkg/defaults/defaults.go
bbguimaraes: Changed.

bbguimaraes:
@stevekuznetsov: added validation for the new test types. This is now ready.

/hold cancel

PR #126, opened 2018-09-04
Add a prototype release creation step

smarterclayton:
The step reuses the pod test logic (factored out) to run a cli command
to craft a new release from the latest stable stream in a PR. The release
payload is published to "release:latest" in the image stream. The step
is always run as long as there is a release tag specification. The step
does nothing if no image "cluster-version-operator" is present in the stable
image stream.

Templates can reference this via `RELEASE_IMAGE_LATEST`.  It is currently an optional target and only a few jobs will directly target it.

Design doc: https://docs.google.com/document/d/1CGZVEyuloZ9oD4NUArW6dnEpi0WFc6BP2tqPSwFZTCY/edit#heading=h.1zgrwxmpgxbr

### DIFF
--- a/pkg/steps/release/create_release.go
+++ b/pkg/steps/release/create_release.go
@@ -1,0 +1,162 @@
+package release
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	imageapi "github.com/openshift/api/image/v1"
+	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	rbacclientset "k8s.io/client-go/kubernetes/typed/rbac/v1"
+
+	"github.com/openshift/ci-operator/pkg/api"
+	"github.com/openshift/ci-operator/pkg/steps"
+)
+
+// assembleReleaseStep knows how to build an update payload image for
+// an OpenShift release by waiting for the full release image set to be
+// created, then invoking the admin command for building a new release.
+type assembleReleaseStep struct {
+	config      api.ReleaseTagConfiguration
+	imageClient imageclientset.ImageV1Interface
+	podClient   steps.PodClient
+	rbacClient  rbacclientset.RbacV1Interface
+	artifactDir string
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
Is there a declarative design doc for the format of the release?

pkg/steps/release/create_release.go
```diff
@@ -0,0 +1,139 @@
...
+	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	rbacclientset "k8s.io/client-go/kubernetes/typed/rbac/v1"
+
+	"github.com/openshift/ci-operator/pkg/api"
+	"github.com/openshift/ci-operator/pkg/steps"
+)
+
+// assembleReleaseStep will tag a full release suite
```
stevekuznetsov: bad doc

pkg/steps/release/create_release.go
```diff
@@ -0,0 +1,139 @@
...
+	release, err := s.imageClient.ImageStreams(s.jobSpec.Namespace).Create(&imageapi.ImageStream{
+		ObjectMeta: meta.ObjectMeta{
+			Name: "release",
+		},
+	})
+	if err != nil {
+		if !errors.IsAlreadyExists(err) {
+			return err
+		}
+		release, err = s.imageClient.ImageStreams(s.jobSpec.Namespace).Get("release", meta.GetOptions{})
```
stevekuznetsov: failed create --> eat error --> try a get

why? this "works" when the create error is a permissions error and the target is already exists?

smarterclayton:
This is still opt in - once this merges I'm going to have the origin GCP jobs depend on this (which will ensure it is build).  Ideally, any repo that is an operator (builds an image with label "io.openshift.release.operator") would have this task run on it, but that can be a later step.

smarterclayton:
Updated as per Steves comments offline, fixed a bug where we were tracking the wrong images (we were using the release tag config, not the stable image stream).

smarterclayton:
Got verbal from Steve, merging so we can start testing this out in the jobs.

PR #129, opened 2018-09-05
Trim labels so they are validated

smarterclayton:
This means we can lose data, but the alternative is a build failure

Fixes https://openshift-gce-devel.appspot.com/build/origin-ci-test/logs/branch-ci-openshift-cluster-api-provider-aws-images-release-3.11/2


smarterclayton:
@stevekuznetsov if you want a different solution just let me know

PR #143, opened 2018-09-14
Config validation updates.

droslean:
Add a validator for the test configuration files.

#### TODOs
- [X] Validate build_root
- [X] Validate base_images
- [X] Validate base_rpm_images
- [X] Validate tests
- [X] Validate promotion
- [X] Validate tag_specification
- [X] Tests cases
- [X] Check for Tests.As duplicates.

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -3,71 +3,238 @@ package api
 import (
 	"errors"
 	"fmt"
+	"net/url"
+	"regexp"
+	"strings"
 
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
-// Validate validates config
+// Validate validates all the configuration's values.
 func (config *ReleaseBuildConfiguration) Validate() error {
-	errs := []error{}
-	if config.Tests != nil {
-		for _, test := range config.Tests {
-			if test.As == "images" {
-				errs = append(errs, errors.New("test should not be called 'images' because it gets confused with '[images]' target"))
-			}
-			typeCount := 0
-			if testConfig := test.ContainerTestConfiguration; testConfig != nil {
-				typeCount += 1
-				// TODO remove when the migration is completed
-				if len(testConfig.From) == 0 && len(test.From) == 0 {
-					errs = append(errs, fmt.Errorf("test %q: `from` is required", test.As))
-				}
-			}
-			if testConfig := test.OpenshiftAnsibleClusterTestConfiguration; testConfig != nil {
-				typeCount += 1
-				errs = append(errs, validateTargetCloud(test.As, testConfig.TargetCloud))
-			}
-			if testConfig := test.OpenshiftAnsibleSrcClusterTestConfiguration; testConfig != nil {
-				typeCount += 1
-				errs = append(errs, validateTargetCloud(test.As, testConfig.TargetCloud))
-			}
-			if testConfig := test.OpenshiftInstallerClusterTestConfiguration; testConfig != nil {
-				typeCount += 1
-				errs = append(errs, validateTargetCloud(test.As, testConfig.TargetCloud))
-			}
-			if testConfig := test.OpenshiftInstallerSmokeClusterTestConfiguration; testConfig != nil {
-				typeCount += 1
-				errs = append(errs, validateTargetCloud(test.As, testConfig.TargetCloud))
-			}
-			if typeCount == 0 {
-				// TODO remove when the migration is completed
-				if len(test.From) == 0 {
-					errs = append(errs, fmt.Errorf("test %q has no type", test.As))
-				}
-			} else if typeCount == 1 {
-				// TODO remove when the migration is completed
-				if len(test.From) > 0 {
-					errs = append(errs, fmt.Errorf("test %q specifies both `From` and a test type", test.As))
-				}
-			} else if typeCount > 1 {
-				errs = append(errs, fmt.Errorf("test %q has more than one type", test.As))
-			}
-		}
+	var validationErrors []error
+
+	validationErrors = append(validationErrors, validateReleaseBuildConfiguration(config))
+	validationErrors = append(validationErrors, validateBuildRootImageConfiguration("build_root", config.InputConfiguration.BuildRootImage))
+	validationErrors = append(validationErrors, validateTestStepConfiguration("tests", config.Tests))
+
+	if config.InputConfiguration.BaseImages != nil {
+		validationErrors = append(validationErrors, validateImageStreamTagReferenceMap("base_images", config.InputConfiguration.BaseImages))
 	}
-	buildRootImage := config.InputConfiguration.BuildRootImage
-	if buildRootImage != nil {
-		if buildRootImage.ProjectImageBuild != nil && buildRootImage.ImageStreamTagReference != nil {
-			errs = append(errs, errors.New("both project_image and image_stream_tag cannot be set for the build_root"))
-		} else if buildRootImage.ProjectImageBuild == nil && buildRootImage.ImageStreamTagReference == nil {
-			errs = append(errs, errors.New("you have to specify either project_image or image_stream_tag for the build_root"))
-		}
+
+	if config.InputConfiguration.BaseRPMImages != nil {
+		validationErrors = append(validationErrors, validateImageStreamTagReferenceMap("base_rpm_images", config.InputConfiguration.BaseRPMImages))
 	}
-	return kerrors.NewAggregate(errs)
+
+	// Validate tag_specification
+	if config.InputConfiguration.ReleaseTagConfiguration != nil {
+		validationErrors = append(validationErrors, validateReleaseTagConfiguration("tag_specification", *config.InputConfiguration.ReleaseTagConfiguration))
+	}
+
+	// Validate promotion in case of `tag_specification` exists or not
+	if config.PromotionConfiguration != nil && config.InputConfiguration.ReleaseTagConfiguration != nil {
+		validationErrors = append(validationErrors, validatePromotionWithTagSpec(*config.PromotionConfiguration, *config.InputConfiguration.ReleaseTagConfiguration))
+	} else if config.PromotionConfiguration != nil && config.InputConfiguration.ReleaseTagConfiguration == nil {
+		validationErrors = append(validationErrors, validatePromotionConfiguration("promotion", *config.PromotionConfiguration))
+	}
+
+	return kerrors.NewAggregate(validationErrors)
 }
 
-func validateTargetCloud(test string, tc TargetCloud) error {
-	if tc != TargetCloudAWS && tc != TargetCloudGCP {
-		return fmt.Errorf("test %q: invalid target cloud %q", test, tc)
+func validatePromotionWithTagSpec(promotion PromotionConfiguration, tagSpec ReleaseTagConfiguration) error {
+	var validationErrors []error
+
+	if len(promotion.Namespace) == 0 && len(tagSpec.Namespace) == 0 {
+		validationErrors = append(validationErrors, fmt.Errorf("promotion: no namespace defined"))
+	}
+	if len(promotion.Name) == 0 && len(promotion.Tag) == 0 {
+		if len(tagSpec.Name) != 0 || len(tagSpec.Tag) != 0 {
+			// will get defaulted, is ok
+		} else {
+			validationErrors = append(validationErrors, errors.New("promotion: no name or tag provided and could not derive defaults from tag_specification"))
+		}
+	}
+
+	return kerrors.NewAggregate(validationErrors)
+}
+
+func validateBuildRootImageConfiguration(fieldRoot string, input *BuildRootImageConfiguration) error {
+	if input == nil {
+		return errors.New("`build_root` is required")
+	}
+
+	if input.ProjectImageBuild != nil && input.ImageStreamTagReference != nil {
+		return fmt.Errorf("%s: both image_stream_tag and project_image_build cannot be set", fieldRoot)
+	} else if input.ProjectImageBuild == nil && input.ImageStreamTagReference == nil {
+		return fmt.Errorf("%s: you have to specify either image_stream_tag or project_image_build", fieldRoot)
 	}
 	return nil
+}
+
+func validateTestStepConfiguration(fieldRoot string, input []TestStepConfiguration) error {
+	var validationErrors []error
+
+	// check for test.As duplicates
+	validationErrors = append(validationErrors, searchForTestDuplicates(input))
+
+	for num, test := range input {
+		if len(test.As) == 0 {
+			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].as: is required", fieldRoot, num))
+		} else if test.As == "images" {
+			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].as: should not be called 'images' because it gets confused with '[images]' target", fieldRoot, num))
+		} else if ok := regexp.MustCompile("^[a-zA-Z0-9_.-]*$").MatchString(test.As); !ok {
+			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].as: `%s` is not valid value, should be [a-zA-Z0-9_.-]", fieldRoot, num, test.As))
+		}
+
+		if len(test.Commands) == 0 {
+			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].commands: is required", fieldRoot, num))
+		}
+
+		validationErrors = append(validationErrors, validateTestConfigurationType(fmt.Sprintf("%s[%d]", fieldRoot, num), test))
+	}
+	return kerrors.NewAggregate(validationErrors)
+}
+
+func validateImageStreamTagReference(fieldRoot string, input ImageStreamTagReference) error {
+	var validationErrors []error
+
+	if _, err := url.ParseRequestURI(input.Cluster); err != nil && len(input.Cluster) != 0 {
+		validationErrors = append(validationErrors, fmt.Errorf("%s.cluster invalid URL given: %s", fieldRoot, err))
+	}
+
+	if len(input.Tag) == 0 {
+		validationErrors = append(validationErrors, fmt.Errorf("%s.tag: value required but not provided", fieldRoot))
+	}
+
+	return kerrors.NewAggregate(validationErrors)
+}
+
+func validateImageStreamTagReferenceMap(fieldRoot string, input map[string]ImageStreamTagReference) error {
+	var validationErrors []error
+	for k, v := range input {
+		if k == "root" {
+			validationErrors = append(validationErrors, fmt.Errorf("%s.%s can't be named `root`", fieldRoot, k))
+		}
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
To summarize the comments I had on Slack for this approach:

Facts:
 - the entire Kubernetes ecosystem does not need anything more complicated than `if` statements for validation
 - Go as a language values simplicity and readability over most everything else
 - this PR adds over ONE MILLION lines of code
 - I do not understand how the validator works and I do not have the time or mental capacity to understand the new dependency

In order to effect the validations that you have been able to add to the fields, we would need:

```go
func validateImageStreamTagReference(fieldRoot string, input ImageStreamTagReference) error {
    var validationErrors []error
    _, err := url.Parse(input.Cluster)
    validationErrors = append(validationErrors, fmt.Errorf("%s.cluster: invalid URL given: %v", fieldRoot, err))
    if input.Tag == "" {
        validationErrors = append(validationErrors, fmt.Errorf("%s.tag: value required but not provided", fieldRoot))
    }
    return kerrors.NewAggregate(validationErrors)
}
```

This code is:
 - not one million lines long
 - apart from the (common) K8S error aggregate, uses _only_ the stdlib
 - simple to understand
 - easy to read
 - straightforward to extend

Let's worry about the actual validation logic and keep things simple.

pkg/api/config.go
```diff
@@ -1,27 +1,199 @@
...
+	return kerrors.NewAggregate(validationErrors)
+}
+
+// Validate validates all the configuration's values.
+func (config *ReleaseBuildConfiguration) Validate() error {
+	if err := deepValidate(reflect.ValueOf(config), []reflect.StructTag{}); err != nil {
+		return err
 	}
 	return nil
 }
```
bbguimaraes: The individual struct validation functions below look good but an `if`-based approach would be more straight-forward in this case than using reflection, as @coverprice mentioned.

This is [how kubernetes does validation](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go).

pkg/api/config_test.go
```diff
@@ -82,7 +99,24 @@ func TestValidate(t *testing.T) {
...
+							Tag: "test",
+						},
+					},
+					BaseImages: map[string]ImageStreamTagReference{"test": {Cluster: "test"},
+						"test2": {Tag: "test2"}, "test3": {Cluster: "test3"},
+					},
+				},
+			},
+			expectedValid: false,
+			expectedError: "[[base_images.test.cluster: invalid URL given: parse test: invalid URI for request, base_images.test.tag: value required but not provided], [base_images.test3.cluster: invalid URL given: parse test3: invalid URI for request, base_images.test3.tag: value required but not provided]]",
```
bbguimaraes: See @stevekuznetsov's comment on removing the check for the specific error message on [my PR](https://github.com/openshift/ci-operator/pull/125#discussion_r218523080).

pkg/api/config.go
```diff
@@ -1,27 +1,199 @@
...
+		validationErrors = append(validationErrors,
+			validateBuildRootImageConfiguration(fieldRoot, iface.Interface().(BuildRootImageConfiguration)))
+
+	// Validate ImageStreamTagReference
+	case imageStreamTagReferenceType:
+		validationErrors = append(validationErrors,
+			validateImageStreamTagReference(fieldRoot, iface.Interface().(ImageStreamTagReference)))
+	}
+
+	if iface.Kind() == reflect.Struct {
```
smarterclayton: why are we doing reflection here?  I don't see how reflection makes this better at validation (for one thing, I can't look to see whether you are checking structs).

stevekuznetsov:
@droslean will review once you rebase

pkg/api/config.go
```diff
@@ -3,71 +3,177 @@ package api
...
-	errs := []error{}
+	var validationErrors []error
+
+	if config.InputConfiguration.BuildRootImage != nil {
+		validationErrors = append(validationErrors, validateBuildRootImageConfiguration("build_root", *config.InputConfiguration.BuildRootImage))
+	} else {
+		validationErrors = append(validationErrors, errors.New("build_root is required"))
+	}
+
 	if config.Tests != nil {
```
stevekuznetsov: This shouldn't really be allowed either

pkg/api/config.go
```diff
@@ -3,71 +3,177 @@ package api
...
+		return fmt.Errorf("%s: both image_stream_tag and project_image_build cannot be set", fieldRoot)
+	} else if input.ProjectImageBuild == nil && input.ImageStreamTagReference == nil {
+		return fmt.Errorf("%s: you have to specify either image_stream_tag or project_image_build", fieldRoot)
+	}
+	return nil
+}
+
+func validateTestStepConfiguration(fieldRoot string, input []TestStepConfiguration) error {
+	var validationErrors []error
+	typeCount := 0
```
stevekuznetsov: This needs to be per-test, not global

pkg/api/config.go
```diff
@@ -3,71 +3,177 @@ package api
...
+}
+
+func validateImageStreamTagReference(fieldRoot string, input ImageStreamTagReference) error {
+	var validationErrors []error
+
+	if _, err := url.ParseRequestURI(input.Cluster); err != nil && input.Cluster != "" {
+		validationErrors = append(validationErrors, fmt.Errorf("%s.cluster invalid URL given: %s", fieldRoot, err))
+	}
+
+	if input.Tag == "" {
```
stevekuznetsov: So we need `namespace` `name` and `tag` in all cases, right? Do we default anything?

pkg/api/config.go
```diff
@@ -3,71 +3,177 @@ package api
...
+		validationErrors = append(validationErrors, fmt.Errorf("%s.tag: value required but not provided", fieldRoot))
+	}
+
+	return kerrors.NewAggregate(validationErrors)
+}
+
+func validateImageStreamTagReferenceMap(fieldRoot string, input map[string]ImageStreamTagReference) error {
+	var validationErrors []error
+	// map has to be sorted to avoid flakes in tests.
```
stevekuznetsov: If we're not trying to validate a specific error message, is this still true?

pkg/api/config.go
```diff
@@ -3,71 +3,177 @@ package api
...
+	sort.Strings(keys)
+	for _, k := range keys {
+		validationErrors = append(validationErrors, validateImageStreamTagReference(fmt.Sprintf("%s.%s", fieldRoot, k), input[k]))
 	}
-	return kerrors.NewAggregate(errs)
+	return kerrors.NewAggregate(validationErrors)
 }
 
-func validateTargetCloud(test string, tc TargetCloud) error {
+func validatePromotionConfiguration(fieldRoot string, input PromotionConfiguration) error {
```
stevekuznetsov: We always need a cluster and namespace

pkg/api/config.go
```diff
@@ -3,71 +3,177 @@ package api
...
+	if len(input.Name) == 0 && len(input.Tag) == 0 {
+		return fmt.Errorf("%s: no name or tag defined for promotion config", fieldRoot)
+	}
+	return nil
+}
+
+func validateReleaseTagConfiguration(fieldRoot string, input ReleaseTagConfiguration) error {
```
stevekuznetsov: We always need a cluster and namespace

pkg/api/config.go
```diff
@@ -3,71 +3,177 @@ package api
...
+
+func validateTestStepConfiguration(fieldRoot string, input []TestStepConfiguration) error {
+	var validationErrors []error
+	typeCount := 0
+
+	// check for test.As duplicates
+	validationErrors = append(validationErrors, searchForTestDuplicates(input))
+
+	for num, test := range input {
+		if len(test.As) == 0 {
```
stevekuznetsov: We need to validate that this is a valid tag, so maybe just check for `[A-Za-z0-9-]`?

pkg/api/config.go
```diff
@@ -3,71 +3,177 @@ package api
...
+	return nil
+}
+
+func searchForTestDuplicates(tests []TestStepConfiguration) error {
+	duplicates := make(map[string]bool, len(tests))
+	var testNames []string
+
+	for _, test := range tests {
+		_, exist := duplicates[test.As]
+		if exist {
```
stevekuznetsov: In Go it's idiomatic to say:


```go
if _, exist := duplicates[test.As]; exist {
```

pkg/api/config.go
```diff
@@ -3,71 +3,177 @@ package api
...
+
+func validateTargetCloud(fieldRoot string, tc TargetCloud) error {
 	if tc != TargetCloudAWS && tc != TargetCloudGCP {
-		return fmt.Errorf("test %q: invalid target cloud %q", test, tc)
+		return fmt.Errorf("%q: invalid target cloud %q", fieldRoot, tc)
+	}
+	return nil
+}
+
+func searchForTestDuplicates(tests []TestStepConfiguration) error {
```
stevekuznetsov: We also need to ensure that there are no collisions between:
 - input base images
 - input base RPM images
 - build root (users can't set this tag name but we need to make sure they don't set any of the other inputs to be it)

pkg/api/config_test.go
```diff
@@ -20,8 +20,9 @@ func TestValidate(t *testing.T) {
 				InputConfiguration: dummyInput,
```
stevekuznetsov: We should break up TestValidate to one test per validation func we have in the validate.go

pkg/api/config.go
droslean: `namespace` and `name` can be empty. We fill them [here](https://github.com/openshift/ci-operator/blob/master/pkg/defaults/defaults.go#L405-L410)

pkg/api/config.go
bbguimaraes: Is there a reason for `tests` to be an array instead of a map?

pkg/api/config.go
droslean: No, it can be removed. The errorMessage have been removed [here](https://github.com/openshift/ci-operator/commit/ef19d3ab9d5cad8f2199873b85d112a44e2b2aa7#diff-787bc42cef1afcb577450319f808c3b2L12)

pkg/api/config.go
droslean: We would need a follow-up issue for that.

pkg/api/config.go
```diff
@@ -3,71 +3,218 @@ package api
...
+	// check for test.As duplicates
+	validationErrors = append(validationErrors, searchForTestDuplicates(input))
+
+	for num, test := range input {
+		if len(test.As) == 0 {
+			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].as: is required", fieldRoot, num))
+		} else if test.As == "images" {
+			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].as: should not be called 'images' because it gets confused with '[images]' target", fieldRoot, num))
+		} else if ok := regexp.MustCompile("^[a-zA-Z0-9_.-]*$").MatchString(test.As); !ok {
+			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].as: `%s` is not valid value", fieldRoot, num, test.As))
```
stevekuznetsov: expose the regex it needs to match so the user can edit their field to fix it

pkg/api/config_test.go
```diff
@@ -1,162 +1,242 @@
...
+
+	for _, tc := range testBaseImagesCases {
+		if err := validateImageStreamTagReferenceMap("base_images", tc.baseImages); err != nil && tc.expectedValid {
+			validationErrors = append(validationErrors, parseError(tc.id, err))
+		} else if !tc.expectedValid && err == nil {
+			validationErrors = append(validationErrors, parseValidError(tc.id))
+		}
+	}
+
+	for _, tc := range testBaseRpmImagesCases {
```
stevekuznetsov: Minor -- let's break this into separate test functions entirely

stevekuznetsov:
```
$ ./ci-operator-checkconfig --config-dir /home/skuznets/releasego/src/github.com/openshift/release/ci-operator/config
validated configuration at containers/libpod/master.yaml
validated configuration at openshift/api/master.yaml
validated configuration at openshift/api/release-3.11.yaml
validated configuration at openshift/azure-misc/master.yaml
validated configuration at openshift/builder/master.yaml
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7fc612]

goroutine 1 [running]:
github.com/openshift/ci-operator/pkg/api.(*ReleaseBuildConfiguration).Validate(0xc420227380, 0x46a, 0x66a)
	/home/skuznets/ci-operator/src/github.com/openshift/ci-operator/pkg/api/config.go:30 +0x402
main.main.func1(0xc42008bab0, 0x6e, 0x96c1e0, 0xc4202271e0, 0x0, 0x0, 0xb, 0xc4204d5bd8)
	/home/skuznets/ci-operator/src/github.com/openshift/ci-operator/cmd/ci-operator-checkconfig/main.go:47 +0x658
path/filepath.walk(0xc42008bab0, 0x6e, 0x96c1e0, 0xc4202271e0, 0xc4203284d0, 0x0, 0x0)
	/usr/local/go/src/path/filepath/path.go:357 +0x402
path/filepath.walk(0xc42008b880, 0x62, 0x96c1e0, 0xc420227040, 0xc4203284d0, 0x0, 0x0)
	/usr/local/go/src/path/filepath/path.go:381 +0x2c2
path/filepath.walk(0xc4200a4a80, 0x56, 0x96c1e0, 0xc420523790, 0xc4203284d0, 0x0, 0x0)
	/usr/local/go/src/path/filepath/path.go:381 +0x2c2
path/filepath.walk(0x7fff19b21d2f, 0x4c, 0x96c1e0, 0xc420522750, 0xc4203284d0, 0x0, 0x10)
	/usr/local/go/src/path/filepath/path.go:381 +0x2c2
path/filepath.Walk(0x7fff19b21d2f, 0x4c, 0xc4203284d0, 0x0, 0x0)
	/usr/local/go/src/path/filepath/path.go:403 +0x106
main.main()
	/home/skuznets/ci-operator/src/github.com/openshift/ci-operator/cmd/ci-operator-checkconfig/main.go:25 +0xe1
```

pkg/api/config.go
```diff
@@ -3,71 +3,235 @@ package api
...
+	if config.InputConfiguration.BaseImages != nil {
+		validationErrors = append(validationErrors, validateImageStreamTagReferenceMap("base_images", config.InputConfiguration.BaseImages))
+	}
+
+	if config.InputConfiguration.BaseRPMImages != nil {
+		validationErrors = append(validationErrors, validateImageStreamTagReferenceMap("base_rpm_images", config.InputConfiguration.BaseRPMImages))
+	}
+
+	if config.PromotionConfiguration != nil && config.ReleaseTagConfiguration != nil {
+
```
stevekuznetsov: If we don't do anything let's drop this case?

pkg/api/config.go
```diff
@@ -3,71 +3,235 @@ package api
...
+		validationErrors = append(validationErrors, fmt.Errorf("%s: no namespace defined", fieldRoot))
+	}
+
+	if len(input.Name) == 0 && len(input.Tag) == 0 {
+		validationErrors = append(validationErrors, fmt.Errorf("%s: no name or tag defined for promotion config", fieldRoot))
+	}
+	return kerrors.NewAggregate(validationErrors)
+}
+
+func validateReleaseTagConfigurationWithoutPromotion(fieldRoot string, input ReleaseTagConfiguration) error {
```
stevekuznetsov: Don't we still need `cluster` and `namespace` here?

pkg/api/config.go
```diff
@@ -3,71 +3,235 @@ package api
...
-	return kerrors.NewAggregate(errs)
+
+	if len(input.Name) == 0 && len(input.Tag) == 0 {
+		validationErrors = append(validationErrors, fmt.Errorf("%s: no name or tag defined", fieldRoot))
+	}
+	return kerrors.NewAggregate(validationErrors)
 }
 
-func validateTargetCloud(test string, tc TargetCloud) error {
+func validateReleaseTagConfiguration(fieldRoot string, input ReleaseTagConfiguration) error {
```
stevekuznetsov: This isn't using `promotion` either?

pkg/api/config.go
```diff
@@ -3,71 +3,235 @@ package api
...
+	} else if config.PromotionConfiguration == nil && config.ReleaseTagConfiguration != nil {
+		validationErrors = append(validationErrors, validateReleaseTagConfigurationWithoutPromotion("tag_specification", *config.InputConfiguration.ReleaseTagConfiguration))
+	} else if config.PromotionConfiguration != nil && config.ReleaseTagConfiguration == nil {
+		validationErrors = append(validationErrors, validatePromotionConfiguration("promotion", *config.PromotionConfiguration))
+	}
+
+	return kerrors.NewAggregate(validationErrors)
+}
+
+func validatePromotionWithTagSpec(promotion PromotionConfiguration, tagSpec ReleaseTagConfiguration) error {
```
stevekuznetsov: Is this ever called?

pkg/api/config.go
```diff
@@ -3,71 +3,230 @@ package api
...
+	var validationErrors []error
+
+	if len(promotion.Name) == 0 && len(promotion.Tag) == 0 {
+		if err := validateReleaseTagConfiguration("tag_specification", tagSpec); err != nil {
+			validationErrors = append(validationErrors, fmt.Errorf("promotion: %v", err))
```
stevekuznetsov: We should just write out the checks -- because today I think the error a user is going to get is confusing.


```go
func validatePromotionWithTagSpec(promotion PromotionConfiguration, tagSpec ReleaseTagConfiguration) error {
	var validationErrors []error

	if len(promotion.Name) == 0 && len(promotion.Tag) == 0 {
		if len(tagSpec.Name) != 0 || len(tagSpec.Tag) != 0 {
			// will get defaulted, is ok
		} else {
			validationErrors = append(validationErrors, errors.New("promotion: no name or tag provided and could not derive defaults from tag_specification"))
		}
	} else {
		validationErrors = append(validationErrors, validatePromotionConfiguration("promotion", promotion))
	}

	return kerrors.NewAggregate(validationErrors)
}
```

pkg/api/config.go
```diff
@@ -3,71 +3,228 @@ package api
...
+	} else if config.PromotionConfiguration == nil && config.InputConfiguration.ReleaseTagConfiguration != nil {
+		validationErrors = append(validationErrors, validateReleaseTagConfiguration("tag_specification", *config.InputConfiguration.ReleaseTagConfiguration))
+	} else if config.PromotionConfiguration != nil && config.InputConfiguration.ReleaseTagConfiguration == nil {
+		validationErrors = append(validationErrors, validatePromotionConfiguration("promotion", *config.PromotionConfiguration))
+	}
+
+	return kerrors.NewAggregate(validationErrors)
+}
+
+func validatePromotionWithTagSpec(promotion PromotionConfiguration, tagSpec ReleaseTagConfiguration) error {
```
stevekuznetsov: So this is still wrong....

If we have both, both need to be validated. The release tag needs to be validated, full stop. The promotion needs to be validated only if the name or tag is specified

PR #148, opened 2018-09-14
Move to using deployments for RPM serving

smarterclayton:
Won't block if a deployment pod gets disrupted

https://jira.coreos.com/browse/DPTP-168


smarterclayton:
Tested manually, might flake if the condition logic isn't perfect but it worked fine in several local runs of repos.

smarterclayton:
@stevekuznetsov

stevekuznetsov:
Nice

/lgtm

PR #154, opened 2018-09-19
remove test_base_image.

droslean:


bbguimaraes:
Reminder that this needs to be communicated because it's a user-facing change. Otherwise…

/lgtm

PR #156, opened 2018-09-19
Set GOPATH when cloning

stevekuznetsov:

/assign @smarterclayton


smarterclayton:
/lgtm

although this is something we shouldn't have to be doing because the image should be defining a GOPATH that we respect.  But it doesn't matter, because GOPATH is going away in Go 1.11

PR #161, opened 2018-09-24
Retry conflicts on promotion

smarterclayton:
Frequent writes to the same namespace can cause conflicts. Retry a few times.


smarterclayton:
Tested manually against a build.

PR #162, opened 2018-09-24
ci-operator should expose RPM_REPO_<org>_<repo>

smarterclayton:
This allows the correct variable to be consumed from templates and
removes ambiguity from use of RPM_REPO. After this change, templates
that need a particular set of RPMs (like ansible installs) ask only for
RPM_REPO_OPENSHIFT_ORIGIN during their template run.

Allows us to complete the move started in 
https://github.com/openshift/release/pull/1645 and remove the `if` check on the new name.

@stevekuznetsov

### DIFF
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -36,10 +36,11 @@ images are built.
 | Example: | `registry.svc.ci.openshift.org/ci-op-<input-hash>/stable:${component}` |
 | - | - |
 
-#### `RPM_REPO`
+#### `RPM_REPO_<org>_<repo>`
 The RPM repository URL from which your job's RPMs can be installed.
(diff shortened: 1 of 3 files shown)


smarterclayton:
/hold

I have to make a fix to how jobs are replaced.

smarterclayton:
I removed RPM_REPO and left only the repo specific version.  Must be merged at the same time as https://github.com/openshift/release/pull/1681

smarterclayton:
Hold removed, review please.

smarterclayton:
Hrm, I wonder if there is something wrong with the centos image

bbguimaraes:
Strange, I just ran the same job on my machine and on `api.ci` and both completed successfully.

/retest

PR #166, opened 2018-10-01
validations doc

droslean:
Add document that describe the validation which will take place while the ci-operator is parsing the configuration file.

### DIFF
--- a/VALIDATIONS.md
+++ b/VALIDATIONS.md
@@ -1,0 +1,36 @@
+# CI Operator Validation Reference
+
+CI Operator will validate the configuration file with the following rules.
+
+
+## General
+* both `tests` and `images` cannot be empty.
+* `rpm_build_location` has to be defined with `rpm_build_commands`.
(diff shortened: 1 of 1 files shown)


VALIDATIONS.md
```diff
@@ -0,0 +1,36 @@
+# CI Operator Validation Reference
```
petr-muller: I was told to always call the thing "ci-operator", not "CI Operator" or something similar

VALIDATIONS.md
```diff
@@ -0,0 +1,36 @@
...
+
+CI-Operator will validate the configuration file with the following rules.
```
petr-muller: Here too

VALIDATIONS.md
```diff
@@ -0,0 +1,36 @@
+# ci-operator Validation Reference
+
+ci-operator will validate the configuration file with the following rules.
```
stevekuznetsov: Nit: we either use "CI Operator" or `ci-operator`.

PR #169, opened 2018-10-03
Ensure e2e tests have correct release

bbguimaraes:
/assign @stevekuznetsov
/cc @droslean @petr-muller

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -10,13 +10,15 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
+var originReleaseTagRegexp = regexp.MustCompile(`^origin-v\d+\.\d+$`)
+
 // Validate validates all the configuration's values.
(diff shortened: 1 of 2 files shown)


pkg/api/config.go
```diff
@@ -210,6 +216,19 @@ func validateTestConfigurationType(fieldRoot string, test TestStepConfiguration)
...
+		if needsReleaseRpms {
+			var version []int
+			if release != nil {
+				if match := originReleaseTagRegexp.FindStringSubmatch(release.Name); match != nil {
+					version = make([]int, 2)
+					version[0], _ = strconv.Atoi(match[1])
+					version[1], _ = strconv.Atoi(match[2])
+				}
+			}
+			if version == nil {
```
droslean: since you are checking if there is no version, why you populate the []int above?
Can you append the error in `if match := originReleaseTagRegexp.FindStringSubmatch(release.Name); match == nil {` ?

PR #170, opened 2018-10-03
The release images step wasn't being added to the dependency map

smarterclayton:
RELEASE_IMAGE_LATEST was not available as parameter even though the task
was conditionally run. Prevented templates from requiring it.

Found when we tried to depend on it in the installer https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_installer/406/pull-ci-openshift-installer-master-e2e-aws/380


smarterclayton:
Fixed https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_installer/406/pull-ci-openshift-installer-master-e2e-aws/381 

verifying it didn't regress others

PR #177, opened 2018-10-11
add active annotation

droslean:
In each creation or resolving a namespace, this annotation will be updated with the current time.
We want to make sure that the namespace will be marked as active in the reuse, to avoid bugs like [#1478](https://github.com/openshift/release/issues/1478)

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -593,6 +593,10 @@ func (o *options) initializeNamespace() error {
 		updates["ci.openshift.io/ttl.hard"] = o.cleanupDuration.String()
 	}
 
+	// This label makes sure that the namespace is active, and the value will be updated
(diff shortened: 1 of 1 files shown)


cmd/ci-operator/main.go
```diff
@@ -593,6 +593,10 @@ func (o *options) initializeNamespace() error {
 		updates["ci.openshift.io/ttl.hard"] = o.cleanupDuration.String()
 	}
 
+	// This label makes sure that the namespace is active, and the value will be updated
+	// if the namespace will be reused.
+	updates["ci.openshift.io/active"] = time.Now().Format(time.RFC3339)
```
stevekuznetsov: The error messages below need to be updated since TTL is not the only thing being updated anymore

PR #179, opened 2018-10-11
Wait until pods are deleted in the initial cleanup

petr-muller:
Fixes: https://github.com/openshift/release/issues/1335 / DPTP-185

To clean up residual template instances and their pods from a possible
previous execution of ci-operator, `template.go:Run()` deletes the
template instance. The comment in the code indicates this does not block
on cleaning up template instance's `Pods`, so the cleanup code waits on
these being deleted after it deletes the template instance.

Previously, the cleanup waited on Pod completion, reusing the code being
used *after* execution to properly collect logs and wait until
everything shuts down gracefully. For some failure cases, this led to
confusing log output being printed during the cleanup, as well as ending
cleanup before the `Pods` were completely deleted. In such case, the
following template instance creation failed because pods to be created
still existed from previous execution. This change makes the cleanup
wait until the `Pods` are actually deleted, and also does not call any
of the code waiting on proper finish and colleting the log output
(because this output belongs to the previous exection anyway).

### DIFF
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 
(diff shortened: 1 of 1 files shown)


pkg/steps/template.go
```diff
@@ -475,7 +476,7 @@ func waitForCompletedTemplateInstanceDeletion(templateClient templateclientset.T
 	for _, ref := range instance.Status.Objects {
 		switch {
 		case ref.Ref.Kind == "Pod" && ref.Ref.APIVersion == "v1":
-			waitForPodCompletion(podClient, ref.Ref.Name, nil)
+			waitForPodDeletion(podClient, ref.Ref.Name, "")
```
stevekuznetsov: Should pass `ref.Ref.UID`?

petr-muller: Right.

petr-muller:
/hold

I'll add a condition to eventually finish the wait loop even if the deletion does not succeed for some reason and also propagate and fail on errors (or at the minimum, log them).

petr-muller:
or not

/meow

PR #182, opened 2018-10-12
Add better validation for resource requests and limits

stevekuznetsov:

/assign @droslean @bbguimaraes @petr-muller

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/api/resource"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/api/config.go
```diff
@@ -229,8 +230,60 @@ func validateReleaseBuildConfiguration(input *ReleaseBuildConfiguration) error {
...
+				}
+				if quantity.Sign() == -1 {
+					validationErrors = append(validationErrors, fmt.Errorf("%s.%s: quantity cannot be negative", fieldRoot, key))
+				}
+			}
+		default:
+			numInvalid++
+			validationErrors = append(validationErrors, fmt.Errorf("`%s` specifies an invalid key %s", fieldRoot, key))
+		}
+	}
```
petr-muller: -paranoid: one thing this won't validate is having a key specified multiple times. Not sure how serious that may be, but hey, we had `resources.resource.resources` :)

pkg/api/config.go
```diff
@@ -229,8 +230,60 @@ func validateReleaseBuildConfiguration(input *ReleaseBuildConfiguration) error {
...
+}
+
+func validateResourceList(fieldRoot string, list ResourceList) error {
+	var validationErrors []error
+
+	var numInvalid int
+	for key := range list {
+		switch key {
+		case "cpu", "memory":
+			if quantity, err := resource.ParseQuantity(list[key]); err != nil {
```
droslean: Do we need to validate a minimum value? For example, in this case, if the user will put `5` instead of `5Gi` we will assume that the value is only 5 bytes. 
Also if the user will input `5Gb` instead of `5Gi` the value will be `0` which is validated below. Do we need to validate also the `Ki | Mi | Gi | Ti | Pi | Ei ` formatting?

pkg/api/config.go
stevekuznetsov: I think it will catch that specific case since the top-level `resources` had no `"*"` entry. Do you mean

```yaml
resources:
  key: ...
  key: ...
```

?

That's impossible to detect with our deserializer

pkg/api/config.go
stevekuznetsov: > Do we need to validate a minimum value? 

I don't think it's super useful. If the user wants to shoot themselves in the foot (in a way that won't crash the ci-operator) so be it ...

> Do we need to validate also the Ki | Mi | Gi | Ti | Pi | Ei formatting?

The parser will do it

stevekuznetsov: > Also if the user will input 5Gb instead of 5Gi the value will be 0 which is validated below

No, that's an invalid suffix

```yaml
resources:
  '*':
    limits:
      cpu: '6'
      memory: 3Gb
    requests:
      cpu: '3'
      memory: 1Gi
```

```
error loading configuration files: invalid configuration from test.yaml: resources.*.limits.memory: invalid quantity: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
```

pkg/api/config.go
petr-muller: Doh. Yeah I meant that but haven't realized we don't see duplicates here already, because we use the same deserializer.

PR #184, opened 2018-10-17
Add annotation to output logs for all containers.

### LINKED ISSUE #118: artifacts: annotation for logs when something else fails
In general, if the e2e tests container fails, we will often want to get the logs from the installer as well, even if the installer container did not fail. We should thread this through as an annotation like `ci-operator.openshift.io/wait-for-container-artifacts` and slap the setup container into there.

/cc @vrutkovs @smarterclayton

droslean:
Adds template's pod annotation `ci-operator.openshift.io/always-show-output`, to output the logs for each container, no matter what the exit code was.

closes #118

### DIFF
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -101,3 +101,13 @@ that exposes a volume named `artifacts`. All files that are added to the volume
 will be uploaded to the appropriate GCS bucket for the job build that executed
 the `Template`. Failures to retrieve or upload artifacts will not impact the
 overall result of the job.
+
+## `Pod`'s annotations
+
+`ci-operator.openshift.io/wait-for-container-artifacts`:
+Will wait for a completion of a container, that has been specified in its value,
+before gathering the artifacts.
+
+`ci-operator.openshift.io/always-show-output`:
+Will output the logs of all the containers in the pod, no matter what the exit code was.
+The value should be `true` to enable this feature.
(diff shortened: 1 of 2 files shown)


bbguimaraes:
As stated multiple times in the issues and in private chat, this needs to be a per-container option. One of the architectural principles of this tool is to be very selective about the logs that are stored (i.e. only those that are strictly needed to determine the cause of failure). This implementation needs to be changed.

droslean:
@bbguimaraes follow up issue opened. #185

smarterclayton:
I don't like this change.  It made debugging failures significantly worse, because I always have to look at output that is irrelevant (e2e, teardown, or setup).  The reason why we didn't show success output is because that makes debugging failures much more efficient (if I have 200k lines of output, I want the last chunk of output to be failures).

I would like to revert this behavior and replace it with something more focused - probably something like "always output logs to the artifacts directory".  A user having to jump into "artifacts" to get at successful container logs is fine for me. Being unable to quickly triage PR flakes and failures by clicking "logs, build logs, end of logs" is not fine.

smarterclayton:
I'm also ok with having the more specific option (store only specific container logs), but I also think it should follow the underlying principle of "show failures and minimal success info in the build log, store anything else in artifacts"

smarterclayton:
An example of where it made success case worse - we had a significant regression in e2e-gcp times - I went to https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/21292/pull-ci-openshift-origin-master-e2e-gcp/891/?log#log to try and see where that is.  Before, I would have ~60 lines and could have easily checked the time of a section.  Now, I might as well give up.

droslean:
@smarterclayton #192

PR #186, opened 2018-10-18
Annotation to force log output containerwise.

### LINKED ISSUE #185: Annotation to force log output containerwise.
Currently, `ci-operator.openshift.io/always-show-output` annotation specifies if the pod will always show the logs no matter what the exit code is. We need to have the same feature, but containerwise.

Add new annotation `ci-operator.openshift.io/always-show-container-output` so specify which container logs `ci-operator` will output after the pod will fail.

droslean:
Change `ci-operator.openshift.io/always-show-output` annotation to specify a list of containers to force the output of the logs.

closes #185

### DIFF
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -111,3 +111,8 @@ before gathering the artifacts.
 `ci-operator.openshift.io/always-show-output`:
 Will output the logs of all the containers in the pod, no matter what the exit code was.
 The value should be `true` to enable this feature.
+
+`ci-operator.openshift.io/always-show-container-output`:
+Will output the logs of a list of containers in the pod, no matter what the exit code was.
+To specify multiple containers the value should be split by `,`. example: `setup,teardown,test`
+This annotation will be overridden if the `ci-operator.openshift.io/always-show-output` is `true`.
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -103,16 +103,12 @@ the `Template`. Failures to retrieve or upload artifacts will not impact the
 overall result of the job.
 
 ## `Pod`'s annotations
+* To specify multiple containers, the value should be split by `,`. example: `"setup,teardown"`
 
 `ci-operator.openshift.io/wait-for-container-artifacts`:
 Will wait for a completion of a container, that has been specified in its value,
 before gathering the artifacts.
 
 `ci-operator.openshift.io/always-show-output`:
-Will output the logs of all the containers in the pod, no matter what the exit code was.
-The value should be `true` to enable this feature.
-
-`ci-operator.openshift.io/always-show-container-output`:
-Will output the logs of a list of containers in the pod, no matter what the exit code was.
-To specify multiple containers the value should be split by `,`. example: `setup,teardown,test`
-This annotation will be overridden if the `ci-operator.openshift.io/always-show-output` is `true`.
+Will output the logs of the specified containers, if the pod fails.
+By default, only the logs of failed containers will be output.
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -38,8 +38,7 @@ type templateExecutionStep struct {
 }
 
 const (
-	showOutputAnnotation          string = "ci-operator.openshift.io/always-show-output"
-	showContainerOutputAnnotation string = "ci-operator.openshift.io/always-show-container-output"
+	showOutputAnnotation string = "ci-operator.openshift.io/always-show-output"
 )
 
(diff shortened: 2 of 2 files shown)


TEMPLATES.md
```diff
@@ -111,3 +111,8 @@ before gathering the artifacts.
 `ci-operator.openshift.io/always-show-output`:
 Will output the logs of all the containers in the pod, no matter what the exit code was.
 The value should be `true` to enable this feature.
+
+`ci-operator.openshift.io/always-show-container-output`:
+Will output the logs of a list of containers in the pod, no matter what the exit code was.
+To specify multiple containers the value should be split by `,`. example: `setup,teardown,test`
```
bbguimaraes: This document is lacking a description of how output from containers is handled by default. We should add that to contrast it with the description of the annotation here.

pkg/steps/template.go
```diff
@@ -38,7 +38,8 @@ type templateExecutionStep struct {
 }
 
 const (
-	showOutputAnnotation string = "ci-operator.openshift.io/always-show-output"
+	showOutputAnnotation          string = "ci-operator.openshift.io/always-show-output"
```
bbguimaraes: This annotation should be removed after we transition all existing users to the new one.

pkg/steps/template.go
```diff
@@ -808,31 +809,43 @@ func podLogNewContainers(podClient coreclientset.PodInterface, pod *coreapi.Pod,
...
-			log.Printf("Container %s in pod %s completed successfully", status.Name, pod.Name)
-			continue
-		}
-
-		if s, err := podClient.GetLogs(pod.Name, &coreapi.PodLogOptions{
-			Container: status.Name,
-		}).Stream(); err == nil {
-			if _, err := io.Copy(os.Stdout, s); err != nil {
-				log.Printf("error: Unable to copy log output from failed pod container %s: %v", status.Name, err)
+		containersToOutput := getContainersMap(strings.Split(pod.ObjectMeta.Annotations[showContainerOutputAnnotation], ","))
```
bbguimaraes: This doesn't have to be done on each iteration of the loop.

pkg/steps/template.go
```diff
@@ -808,31 +809,43 @@ func podLogNewContainers(podClient coreclientset.PodInterface, pod *coreapi.Pod,
...
+		if _, exists := containersToOutput[status.Name]; pod.ObjectMeta.Annotations[showOutputAnnotation] == "true" || s.ExitCode != 0 || exists {
```
bbguimaraes: This does not follow what's defined in the user story:

> I want to see the logs from my cluster setup and teardown pods when the test pod fails

> […] we need to grab the logs from the setup container when the test container fails even if the setup container did not fail.

pkg/steps/template.go
```diff
@@ -809,8 +808,8 @@ func podLogNewContainers(podClient coreclientset.PodInterface, pod *coreapi.Pod,
 		completed[status.Name] = s.FinishedAt.Time
 		notifier.Notify(pod, status.Name)
 
-		containersToOutput := getContainersMap(strings.Split(pod.ObjectMeta.Annotations[showContainerOutputAnnotation], ","))
-		if _, exists := containersToOutput[status.Name]; pod.ObjectMeta.Annotations[showOutputAnnotation] == "true" || s.ExitCode != 0 || exists {
+		containersToOutput := getContainersMap(pod.ObjectMeta.Annotations[showOutputAnnotation])
```
bbguimaraes: This doesn't have to be done on each iteration of the loop.

pkg/steps/template.go
```diff
@@ -809,8 +808,8 @@ func podLogNewContainers(podClient coreclientset.PodInterface, pod *coreapi.Pod,
...
+		if _, exists := containersToOutput[status.Name]; s.ExitCode != 0 || exists {
```
bbguimaraes: This does not follow the description of the issue:

> As a developer on an installer team
> I want to see the logs from my cluster setup and teardown pods when the test pod fails
> […]
> we need to grab the logs from the setup container when the test container fails even if the setup container did not fail.

kargakis:
Showing the output of complete tests is still useful and something we need. 

Example: The following job succeeds but one of the ginkgo tests in the job actually failed which is a bug in our tests. We would never figure this out w/o unconditionally printing the whole log by default.

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/608/pull-ci-openshift-openshift-azure-master-e2e-azure/340?log#log

PR #188, opened 2018-10-19
pkg/steps/project_image: Use org.opencontainers.image labels

wking:
These are [standardized][1], and folks consuming our images are more likely to recognize the standard names than our local labels.

I'm adding a new label for the source repository to make it easier to look up images where the mapping between the image tag and source repo is not obvious (e.g. olm -> [operator-framework/operator-lifecycle-manager][2]).

@smarterclayton mentioned something about these maybe being consumed by an OpenShift UI, in which case we may want to deprecate `io.openshift.build.commit.id` instead of removing it.

CC @RobertKrawitz, who's scraping the commit IDs out of the image metadata to help track the source of various cluster-install issues.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md#pre-defined-annotation-keys
[2]: https://github.com/operator-framework/operator-lifecycle-manager

### DIFF
--- a/pkg/steps/project_image.go
+++ b/pkg/steps/project_image.go
@@ -50,10 +50,12 @@ func (s *projectDirectoryImageBuildStep) Run(ctx context.Context, dry bool) erro
 
 	labels := make(map[string]string)
 	if len(s.jobSpec.Refs.Pulls) == 0 {
-		labels["io.openshift.build.commit.id"] = s.jobSpec.Refs.BaseSHA
+		labels["org.opencontainers.image.source"] = fmt.Sprintf("https://github.com/%s/%s", s.jobSpec.Refs.Org, s.jobSpec.Refs.Repo)
+		labels["org.opencontainers.image.revision"] = s.jobSpec.Refs.BaseSHA
 		labels["io.openshift.build.commit.ref"] = s.jobSpec.Refs.BaseRef
 	} else {
-		labels["io.openshift.build.commit.id"] = ""
+		labels["org.opencontainers.image.source"] = ""
+		labels["org.opencontainers.image.revision"] = ""
 		labels["io.openshift.build.commit.ref"] = ""
 	}
 


pkg/steps/project_image.go
```diff
@@ -50,10 +50,12 @@ func (s *projectDirectoryImageBuildStep) Run(ctx context.Context, dry bool) erro
 
 	labels := make(map[string]string)
 	if len(s.jobSpec.Refs.Pulls) == 0 {
-		labels["io.openshift.build.commit.id"] = s.jobSpec.Refs.BaseSHA
+		labels["org.opencontainers.image.source"] = fmt("https://github.com/%s/%s", s.jobSpec.Refs.Org, s.jobSpec.Refs.Repo)
```
wking: `fmt.Sprintf`

wking: > `fmt.Sprintf`

Fixed in c59664d -> 12badc4.

smarterclayton:
/hold

I need to figure out what we’re doing for actual OSBS images, which set
some of this and have a specific format that we can’t change.  We may just
use that.  Will look Monday.

On Oct 19, 2018, at 5:39 PM, W. Trevor King <notifications@github.com>
wrote:

These are standardized
<https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md#pre-defined-annotation-keys>,
and folks consuming our images are more likely to recognize the standard
names than our local labels.

I'm adding a new label for the source repository to make it easier to look
up images where the mapping between the image tag and source repo is not
obvious (e.g. olm -> operator-framework/operator-lifecycle-manager
<https://github.com/operator-framework/operator-lifecycle-manager>).

@smarterclayton <https://github.com/smarterclayton> mentioned something
about these maybe being consumed by an OpenShift UI, in which case we may
want to deprecate io.openshift.build.commit.id instead of removing it.

CC @RobertKrawitz <https://github.com/RobertKrawitz>, who's scraping the
commit IDs out of the image metadata to help track the source of various
cluster-install issues.
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/openshift/ci-operator/pull/188
Commit Summary

   - pkg/steps/project_image: Use org.opencontainers.image labels

File Changes

   - *M* pkg/steps/project_image.go
   <https://github.com/openshift/ci-operator/pull/188/files#diff-0> (6)

Patch Links:

   - https://github.com/openshift/ci-operator/pull/188.patch
   - https://github.com/openshift/ci-operator/pull/188.diff

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/ci-operator/pull/188>, or mute the thread

pkg/steps/project_image.go
```diff
@@ -50,10 +50,12 @@ func (s *projectDirectoryImageBuildStep) Run(ctx context.Context, dry bool) erro
 
 	labels := make(map[string]string)
 	if len(s.jobSpec.Refs.Pulls) == 0 {
-		labels["io.openshift.build.commit.id"] = s.jobSpec.Refs.BaseSHA
+		labels["org.opencontainers.image.source"] = fmt.Sprintf("https://github.com/%s/%s", s.jobSpec.Refs.Org, s.jobSpec.Refs.Repo)
```
smarterclayton: Should these end in .git?

We need the ref for other reasons.  We probably also need to correctly indicate merge commits.

wking: > Should these end in .git?

You can `pull` either way.  And hitting a`.git` in my browser redirects me to the non-`.git` form, so *shrug*.  I like the way I have it, but will change if you like.

> We need the ref for other reasons.

What ref, the old `io.openshift.build.commit.id`?

> We probably also need to correctly indicate merge commits.

Follow-up work?

wking:
> I need to figure out what we’re doing for actual OSBS image...

Can you point me at the OSBS source.  I've found [the][1] [docs][2], but the source for the project itself is not clear to me.  Maybe the core is [atomic-reactor][3], but it has nothing to say about `io.openshift.`.  [This repo][4] has some documentation for labels, but only [two unrelated labels][5] under `io.openshift.`.  And it's not clear to me who is consuming those labels.

[1]: https://osbs.readthedocs.io/en/latest/
[2]: https://github.com/projectatomic/osbs-docs
[3]: https://github.com/projectatomic/atomic-reactor
[4]: https://github.com/projectatomic/ContainerApplicationGenericLabels
[5]: https://github.com/projectatomic/ContainerApplicationGenericLabels/blame/3e158d906132d157b66e6a9c9630d4298bba3af7/vendor/redhat/labels.md#L47-L48

smarterclayton:
OSBS sets these:

vcs-ref
authoritative-source-url
vcs-type

wking:
> OSBS sets these:
>
> vcs-ref
> authoritative-source-url
> vcs-type

Should I start wading through [these][1] and similar porting them to the OCI tags?

[1]: https://github.com/search?p=3&q=%22authoritative-source-url%22+org%3Aprojectatomic&type=Code

smarterclayton:
Good question.  It really sucks we have three different sets.

smarterclayton:
Nevermind, open container doesn't have branch info.  See #196

PR #189, opened 2018-10-22
output logs containerwise.

### LINKED ISSUE #185: Annotation to force log output containerwise.

droslean:
Adds the `ci-operator.openshift.io/always-show-container-output` annotation to specify which container logs, will be shown **after** the pod will fail.

closes #185

### DIFF
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -105,9 +105,13 @@ overall result of the job.
 ## `Pod`'s annotations
 
 `ci-operator.openshift.io/wait-for-container-artifacts`:
-Will wait for a completion of a container, that has been specified in its value,
-before gathering the artifacts.
+Comma-separated list of container names for which ci-operator will wait until complete,
+to gather the artifacts.
 
 `ci-operator.openshift.io/always-show-output`:
 Will output the logs of all the containers in the pod, no matter what the exit code was.
 The value should be `true` to enable this feature.
+
+`ci-operator.openshift.io/containers-logged-on-failure`:
+Comma-separated list of container names for which ci-operator will collect and log their output if pod fails.
+By default, only the logs from the failed containers will be output.
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -38,7 +38,8 @@ type templateExecutionStep struct {
 }
 
 const (
-	showOutputAnnotation string = "ci-operator.openshift.io/always-show-output"
+	showOutputAnnotation          string = "ci-operator.openshift.io/always-show-output"
+	showContainerOutputAnnotation string = "ci-operator.openshift.io/containers-logged-on-failure"
 )
 
 func (s *templateExecutionStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, error) {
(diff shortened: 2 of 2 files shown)


petr-muller:
I am confused about the intention of this change. I expected to find a different feature here! :)

My understanding is the following:

1. In https://github.com/openshift/ci-operator/pull/184, we added a feature where, providing a boolean annotation, the users can force output from all containers to be present, no matter the exit code.
2. @bbguimaraes [requested](https://github.com/openshift/ci-operator/pull/184#issuecomment-430624540) changes in the feature and you filed https://github.com/openshift/ci-operator/issues/185 for it
3. This PR is supposed to close #185 but to me, it looks this is a different feature.

Can you clarify? How is this related to #185? If it's unrelated, what is this feature (a different issue?).

TEMPLATES.md
```diff
@@ -111,3 +111,10 @@ before gathering the artifacts.
 `ci-operator.openshift.io/always-show-output`:
 Will output the logs of all the containers in the pod, no matter what the exit code was.
 The value should be `true` to enable this feature.
+
+`ci-operator.openshift.io/always-show-container-output`:
```
petr-muller: I'm wondering if `always-show-container-output` is the right name for the annotation (the name was reused from the previous feature, right?). It may be misleading because it looks similar to the `always-show-output` one but works differently. I don't like the `always` part - it's not true (always != when pod fails and only if container terminated) and it is what makes it similar to `always-show-output`.

I am not sure about a good name, though... `containers-logged-on-failure` ?

TEMPLATES.md
```diff
@@ -111,3 +111,10 @@ before gathering the artifacts.
...
+Will output the logs of the specified containers after the pod fails, if
+they are in a terminated state.
```
petr-muller: Briefly describe the default state somewhere. What containers' output is captured by default?

TEMPLATES.md
```diff
@@ -111,3 +111,10 @@ before gathering the artifacts.
...
+
+**Note:**
+To specify multiple containers in `ci-operator.openshift.io/wait-for-container-artifacts` and `ci-operator.openshift.io/always-show-container-output` annotations, the value should be splited by `,`.
```
petr-muller: Perhaps include this into the descriptions themselves? This would require rewritting all of them into a different form, but I think it would be worth the improvement.

_Comma-separated list of container names for which ci-operator will wait until complete / for which ci-operator will collect and log their output if pod fails_

droslean:
@bbguimaraes requested the changes, but in the end, it turned out that we will need the feature described in #184. This PR adds a new feature, described in #185.

pkg/steps/template.go
```diff
@@ -836,3 +837,49 @@ func podLogNewContainers(podClient coreclientset.PodInterface, pod *coreapi.Pod,
...
+		}
+		s.Close()
+	} else {
+		return fmt.Errorf("error: Unable to retrieve logs from failed pod container %s: %v", statusName, err)
+	}
+	return nil
+}
+
+func getContainersMap(containers string) map[string]struct{} {
+	s := strings.Split(containers, ",")
```
petr-muller: I'm thinking of [stripping whitespace](https://golang.org/pkg/strings/#TrimSpace) to be friendly to people specifying the list as `aaa, bbb`

PR #190, opened 2018-10-22
Add `--as` option for impersonating users

petr-muller:
Useful for features for we need `system:admin`, like namespace annotations and/or VMs

/cc @droslean @bbguimaraes @stevekuznetsov


stevekuznetsov:
How is this used?

petr-muller:
> How is this used?

I was using this to test the prototypes of the steps using `ci-vm-operator` to get VMs.

PR #191, opened 2018-10-25
Add an openstack profile

flaper87:


droslean:
Since you are doing this, you should update the validator as well.
check: https://github.com/openshift/ci-operator/blob/master/pkg/api/config.go#L157

/hold

flaper87:
> Since you are doing this, you should update the validator as well.
> check: https://github.com/openshift/ci-operator/blob/master/pkg/api/config.go#L157
> 
> /hold

mmh, unless I'm missing something, I've done that already by adding `ClusterProfileOpenStack` to the list. Is that wrong?

droslean:
@flaper87 The validator won't let you use it.

PR #192, opened 2018-10-26
remove deprecated annotation

droslean:
This annotation is going to be deprecated since it is making debugging more complicated.


smarterclayton:
I'm ok with having an annotation to request artifacts for this container be always stored - that might be a good compromise.  We can discuss more.

PR #193, opened 2018-10-30
Log containers output to artifacts.

### LINKED ISSUE #194: Container logs are printed in a loop when a test fails
We are seeing container logs getting printed in an endless loop in our test logs, eg. [here](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_openshift-azure/670/pull-ci-openshift-openshift-azure-master-e2e-azure/469). Maybe related: [we have chosen to print all container logs when there is a template failure](https://github.com/openshift/release/blob/e274d0cc4d64e55bad5ad13c29ec892f4f273efd/ci-operator/templates/openshift/openshift-azure/cluster-launch-e2e-azure.yaml#L44). 

/kind bug
@droslean @stevekuznetsov

droslean:
Changes the logic of `ci-operator.openshift.io/containers-logged-on-failure` annotation to log the output of the container to artifacts instead of the `ci-operator`'s stdout.

Also fixes #194

### DIFF
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -110,4 +110,4 @@ to gather the artifacts.
 
 `ci-operator.openshift.io/containers-logged-on-failure`:
 Comma-separated list of container names for which ci-operator will collect and log their output to artifacts if pod fails.
-The artifact name will be generated in the format of `$container_name-out.log`.
+The artifact names will be generated in the format of `$container_name.log` in the `container-logs` directory.
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -109,5 +109,5 @@ Comma-separated list of container names for which ci-operator will wait until co
 to gather the artifacts.
 
 `ci-operator.openshift.io/containers-logged-on-failure`:
-Comma-separated list of container names for which ci-operator will collect and log their output if pod fails.
-By default, only the logs from the failed containers will be output.
+Comma-separated list of container names for which ci-operator will collect and log their output to artifacts if pod fails.
+The artifact name will be generated in the format of `$container_name-out.log`.
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -22,6 +22,7 @@ import (
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/watch"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 
@@ -156,7 +157,7 @@ func (s *templateExecutionStep) Run(ctx context.Context, dry bool) error {
 	for _, ref := range instance.Status.Objects {
 		switch {
 		case ref.Ref.Kind == "Pod" && ref.Ref.APIVersion == "v1":
-			if err := waitForPodCompletion(s.podClient.Pods(s.jobSpec.Namespace), ref.Ref.Name, notifier); err != nil {
+			if err := waitForPodCompletion(s.podClient.Pods(s.jobSpec.Namespace), ref.Ref.Name, notifier, s.artifactDir); err != nil {
 				return fmt.Errorf("could not wait for pod to complete: %v", err)
 			}
 		}
@@ -549,7 +550,7 @@ func waitForCompletedPodDeletion(podClient coreclientset.PodInterface, name stri
 	return waitForPodDeletion(podClient, name, uid)
 }
 
-func waitForPodCompletion(podClient coreclientset.PodInterface, name string, notifier ContainerNotifier) error {
+func waitForPodCompletion(podClient coreclientset.PodInterface, name string, notifier ContainerNotifier, artifactDir string) error {
 	if notifier == nil {
 		notifier = NopNotifier
 	}
@@ -564,6 +565,15 @@ func waitForPodCompletion(podClient coreclientset.PodInterface, name string, not
 			continue
 		}
 		if err != nil {
+			if len(artifactDir) > 0 {
+				podList, err := getPodListByName(podClient, name)
+				if err != nil {
+					log.Printf("could not list pod: %v", err)
+				}
+				if err := artifactsAnnotatedContainerLogs(podClient, &podList.Items[0], artifactDir); err != nil {
+					log.Printf("%v", err)
+				}
+			}
 			return fmt.Errorf("could not wait for pod completion: %v", err)
 		}
 		if !retry {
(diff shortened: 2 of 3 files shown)


TEMPLATES.md
```diff
@@ -109,5 +109,5 @@ Comma-separated list of container names for which ci-operator will wait until co
 to gather the artifacts.
 
 `ci-operator.openshift.io/containers-logged-on-failure`:
-Comma-separated list of container names for which ci-operator will collect and log their output if pod fails.
-By default, only the logs from the failed containers will be output.
\ No newline at end of file
+Comma-separated list of container names for which ci-operator will collect and log their output to artifacts if pod fails.
+The artifact name will be generated in the format of `$container_name-out.log`.
```
petr-muller: What is the chance of a name clash with a file that the test will want to put there? This will lead to filenames like `setup-out.log`, not that unique. Should we prefix this, or maybe put it into a separate directory?

pkg/steps/template.go
```diff
@@ -856,20 +861,32 @@ func getContainersMap(containers string) map[string]struct{} {
...
+
 	for _, status := range statuses {
 		if s := status.State.Terminated; s != nil {
 			if _, exists := containersToOutput[status.Name]; exists {
-				log.Printf("Container %s logs:", status.Name)
-				if err := printContainerLogsToStdout(podClient, status.Name, pod.Name); err != nil {
-					return fmt.Errorf("%v", err)
+				file, err := os.Create(fmt.Sprintf("%s/%s/%s-out.log", artifactDir, pod.Name, status.Name))
+				if err != nil {
+					validationErrors = append(validationErrors, fmt.Errorf("Cannot create file: %v", err))
```
petr-muller: Is there any reason for not skipping the rest of the body here? What is the point of continuing?

pkg/steps/template.go
```diff
@@ -856,20 +861,32 @@ func getContainersMap(containers string) map[string]struct{} {
...
 	statuses := gatherContainerStatuses(pod)
 	containersToOutput := getContainersMap(pod.ObjectMeta.Annotations[showContainerOutputAnnotation])
+
 	for _, status := range statuses {
 		if s := status.State.Terminated; s != nil {
 			if _, exists := containersToOutput[status.Name]; exists {
-				log.Printf("Container %s logs:", status.Name)
-				if err := printContainerLogsToStdout(podClient, status.Name, pod.Name); err != nil {
-					return fmt.Errorf("%v", err)
+				file, err := os.Create(fmt.Sprintf("%s/%s/%s-out.log", artifactDir, pod.Name, status.Name))
```
petr-muller: What creates the `pod.Name` subdirectory of artifact dir? Will it always exist? 

Also, this is not consistent with the description in the documentation change - it does not mention any subdirectory.

droslean: It will continue for the rest containers. The subdirectory is already being created with `--artifact-dir`

TEMPLATES.md
droslean: Yes, this is a possibility that I also consider, that's why I wanted to specify the name in the docs. So the user will avoid writing his script to output to a file with the same name. It can be avoided also programmatically, but this will lead in an extra code as well.

petr-muller: If we reserve a filename this way, it should be something more specific. IMHO.

pkg/steps/template.go
petr-muller: Can you provide a link? When is that subdirectory created?

pkg/steps/template.go
petr-muller: Yeah, in the next cycle iteration, but why executing the rest of the body? The `Copy` there will fail no matter what.

pkg/steps/template.go
droslean: https://github.com/openshift/ci-operator/blob/master/pkg/steps/artifacts.go#L117-L121

pkg/steps/template.go
droslean: Yes, the intention here is output the errors if there are also for the other containers.

pkg/steps/template.go
```diff
@@ -620,9 +623,6 @@ func waitForPodCompletionOrTimeout(podClient coreclientset.PodInterface, name st
 				return false, nil
 			}
 			if podJobIsFailed(pod) {
-				if err := outputAnnotatedContainerLogs(podClient, pod); err != nil {
-					log.Printf("%v", err)
-				}
```
petr-muller: Why is the output being removed here? If we miss the first event, removing this would mean we do not log the failed container output, woudn't it?

droslean: It will.  See https://github.com/openshift/ci-operator/pull/193/files#diff-37ed7e4b3e5f7bb5d23118bff1f75bd9R597

pkg/steps/template.go
petr-muller: Maybe I am missing something, but I do not see any code that would create a subdirectory for each container there - it simply extracts everything from a tarball, which is created by tarring a content of the `/tmp/artifacts`.

pkg/steps/template.go
petr-muller: Other containers will be handled by different iterations of the cycle. I am asking why to continue executing a single iteration when a file creation fails.

pkg/steps/template.go
petr-muller: That happens before this removed code, and may not hit (otherwise this cycle would not be needed).

TEMPLATES.md
droslean: Will it be better, if we send them in a different directory, inside pod's directory of course, with the $containername.log format?

pkg/steps/template.go
droslean: @petr-muller Its in https://github.com/openshift/ci-operator/blob/master/pkg/steps/template.go#L136
```go
filepath.Join(s.artifactDir, s.template.Name)
```

pkg/steps/template.go
```diff
@@ -564,6 +565,15 @@ func waitForPodCompletion(podClient coreclientset.PodInterface, name string, not
 			continue
 		}
 		if err != nil {
+			if len(artifactDir) > 0 {
+				podList, err := getPodListByName(podClient, name)
+				if err != nil {
+					log.Printf("could not list pod: %v", err)
```
petr-muller: Cannot continue to execute when `err != nil`, because then `podList` can be `nil` on line 573 and we will SEGV.

TEMPLATES.md
petr-muller: Like having something like `container-logs/`? SGTM

pkg/steps/template.go
petr-muller: This still holds.

petr-muller:
/lgtm

Thanks!

smarterclayton:
This caused https://github.com/openshift/ci-operator/issues/197, which was fixed in #198

PR #195, opened 2018-10-30
remove output of log while waiting.

### LINKED ISSUE #194: Container logs are printed in a loop when a test fails

droslean:
closes #194


droslean:
That will change on #193

petr-muller:
As far as I understand #193 , that should only change a destination where the output is captured.

PR #196, opened 2018-11-01
build: Clear more default annotations when we build images

smarterclayton:
Ensure that the set of annotations on a project image corresponding
to information about the source input are not copied, so that we
can use these downstream in the release tooling.

Start setting vcs-ref/vcs-url/vcs-type like OSBS

https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md

Fixes #188

### DIFF
--- a/pkg/steps/project_image.go
+++ b/pkg/steps/project_image.go
@@ -49,12 +49,33 @@ func (s *projectDirectoryImageBuildStep) Run(ctx context.Context, dry bool) erro
 	}
 
 	labels := make(map[string]string)
+	// reset all labels that may be set by a lower level
+	for _, key := range []string{
+		"vcs-type",
+		"vcs-ref",
+		"vcs-url",
+		"io.openshift.build.name",
+		"io.openshift.build.namespace",
+		"io.openshift.build.commit.id",
+		"io.openshift.build.commit.ref",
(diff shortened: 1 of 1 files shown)


pkg/steps/project_image.go
```diff
@@ -49,12 +49,33 @@ func (s *projectDirectoryImageBuildStep) Run(ctx context.Context, dry bool) erro
 	}
 
 	labels := make(map[string]string)
+	// reset all labels that may be set by a lower level
+	for _, key := range []string{
+		"vcs-type",
+		"vcs-ref",
+		"vcs-url",
```
wking: These non-namespaced labels violate [the SHOULD-level reverse domain recommendation][1].  I'd really like to get these ported to [the OCI standard labels][2], so folks can understand them without knowing about Red-Hat-specific labels.  What do I need to change upstream before we can change these here?  Or can we add the OCI labels in parallel here (I don't care as much if the information is available via non-standard labels as long as the standard labels are there).

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.1/annotations.md#L11
[2]: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md#pre-defined-annotation-keys

Issue #200, opened 2018-11-07
failed test jobs report incorrect error message


bparees:
This job had an (expected) test failure, but instead of reporting that, it reported a confusing top level job failure message of "error: could not run steps: could not wait for pod to complete: could not wait for pod completion: the pod ci-op-dibs34hg/e2e-aws-operator failed after 46m30s (failed containers: test):  unknown"

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-image-registry-operator/59/pull-ci-openshift-cluster-image-registry-operator-master-e2e-aws-operator/2

```
--- FAIL: TestClusterImageRegistryOperator (1.13s)
	cluster_image_registry_operator_test.go:25: unexpected error: imageregistries.imageregistry.operator.openshift.io "fake-image-registry" not found
FAIL
FAIL	github.com/openshift/cluster-image-registry-operator/test/e2e	1.148s
2018/11/07 01:15:52 Container test in pod e2e-aws-operator failed, exit code 1, reason Error
2018/11/07 01:23:49 Container teardown in pod e2e-aws-operator completed successfully
2018/11/07 01:23:49 Copying artifacts from e2e-aws-operator into /logs/artifacts/e2e-aws-operator
2018/11/07 01:23:50 Copied 6.35Mi of artifacts from e2e-aws-operator to /logs/artifacts/e2e-aws-operator
2018/11/07 01:23:55 Ran for 50m2s
error: could not run steps: could not wait for pod to complete: could not wait for pod completion: the pod ci-op-dibs34hg/e2e-aws-operator failed after 46m30s (failed containers: test):  unknown
```

/cc @smarterclayton

droslean:
It reported that the `test` container has failed. You can check why it failed in the above lines since we print the errors of any container.

bparees:
it reports that it could not wait for the pod to complete, which implies something timed out or went fundamentally wrong.   The top level error a user is confronted with when they go to the job url(https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-image-registry-operator/59/pull-ci-openshift-cluster-image-registry-operator-master-e2e-aws-operator/2) is:


"operator Instantiate the template e2e-aws-operator into the operator namespace and wait for any pods to complete 46m38s"

[...]

PR #201, opened 2018-11-08
Fix error messages.

droslean:
closes: #200


pkg/steps/template.go
```diff
@@ -570,7 +570,7 @@ func waitForPodCompletion(podClient coreclientset.PodInterface, name string, not
 					log.Printf("%v", err)
 				}
 			}
-			return fmt.Errorf("could not wait for pod completion: %v", err)
+			return fmt.Errorf("%v", err)
```
petr-muller: Why not returning `err` straight?

```suggestion
			return err
```

Issue #203, opened 2018-11-09
ci-operator stacktrace on artifact publishing?


stevekuznetsov:
```
2018/11/09 15:13:24 Resolved source https://github.com/operator-framework/operator-lifecycle-manager to master@c82dca53, merging: #560 b3f46e52 @njhale
2018/11/09 15:13:24 Resolved openshift/release:golang-1.10 to sha256:571702190389462220e31ea84041b2a53e56141e799a56dac73c39ed90e22c70
2018/11/09 15:13:24 Resolved openshift/origin-v4.0:base to sha256:9f831b0d8d1b50c928b84d87cb59ef7c84c58b41462f5e10ba154d7853c45b3a
2018/11/09 15:13:24 Using namespace ci-op-xwxn9k5k
2018/11/09 15:13:24 Running [input:root], [input:base], [release-inputs], src, olm, [output:stable:olm], [images], [release:latest], e2e-aws-olm
2018/11/09 15:13:24 Creating namespace ci-op-xwxn9k5k
2018/11/09 15:13:24 Creating rolebinding for user njhale in namespace ci-op-xwxn9k5k
2018/11/09 15:13:25 Setting a soft TTL of 1h0m0s for the namespace
2018/11/09 15:13:25 Setting a hard TTL of 12h0m0s for the namespace
2018/11/09 15:13:25 Setting up pipeline imagestream for the test
2018/11/09 15:13:25 Populating secrets for test
2018/11/09 15:13:25 Created secret e2e-aws-olm-cluster-profile
2018/11/09 15:13:25 Tagging openshift/release:golang-1.10 into pipeline:root
2018/11/09 15:13:25 Tagging openshift/origin-v4.0:base into pipeline:base
2018/11/09 15:13:25 Tagged release images from openshift/origin-v4.0:${component}, images will be pullable from registry.svc.ci.openshift.org/ci-op-xwxn9k5k/stable:${component}
2018/11/09 15:13:25 Building src
2018/11/09 15:13:51 Build src succeeded after 25s
2018/11/09 15:13:51 Building olm
2018/11/09 15:16:45 Build olm succeeded after 2m53s
2018/11/09 15:16:45 Tagging olm into stable
2018/11/09 15:16:45 All images ready
2018/11/09 15:16:45 Create a new update payload image registry.svc.ci.openshift.org/ci-op-xwxn9k5k/release:latest
2018/11/09 15:16:45 Executing release release-latest
[...]

PR #205, opened 2018-11-11
Revert container output logging until it can be fixed

smarterclayton:
Until we know how to fix https://github.com/openshift/ci-operator/issues/203 I'm reverting the container output logging commits (described below) back to the first implementation.

In general, this is a responsibility of the notifier, not of waitForPodCompletion. The notifier is the one who should be looking at the annotation and taking action, which
removes the need to modify template.go. By moving the logic there we can avoid the failures we're seeing.

This reverts commit 5016ebed34abf5a35b7ce758f8c18fbe1eea9070.
This reverts commit 30e3e1f43c0c9789c4b754fb9e7293e709b332f9.
This reverts commit bb15ab407343ebb9c80422af755ded041abb1adc.
This reverts commit d8a3422c4ac6358934c5aff82215b83058ec5b6a.
This reverts commit 2ad5a13f9ffce8fcc1c829eafbae7c99fe4ac336.
This reverts commit 09897c9315719e76480a2bdd6725b0471efc1c36.
This reverts commit 75ff81e8b520bc018f80f6d0bcac9b435180b7bd.

### DIFF
--- a/TEMPLATES.md
+++ b/TEMPLATES.md
@@ -101,13 +101,3 @@ that exposes a volume named `artifacts`. All files that are added to the volume
 will be uploaded to the appropriate GCS bucket for the job build that executed
 the `Template`. Failures to retrieve or upload artifacts will not impact the
 overall result of the job.
-
-## `Pod`'s annotations
-
-`ci-operator.openshift.io/wait-for-container-artifacts`:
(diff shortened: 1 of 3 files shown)


smarterclayton:
Please add me as a reviewer on the updated PR implementation using the artifact notifier and I'll help review.  Thanks

PR #206, opened 2018-11-12
Container output to artifacts

droslean:
Adds a way to gather all the container outputs to artifacts by setting an annotation to the pod.

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -520,7 +520,7 @@ func gatherContainerLogsOutput(podClient PodClient, artifactDir, namespace, podN
 	}
 	pod := &list.Items[0]
 
-	if pod.Annotations["ci-operator.openshift.io/container-logs-as-artifacts"] == "" {
+	if pod.Annotations["ci-operator.openshift.io/save-container-logs"] != "true" {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/steps/artifacts.go
```diff
@@ -505,3 +511,65 @@ func allPodContainerNamesWithArtifacts(pod map[string]interface{}) map[string]st
...
+
+func gatherContainerLogsOutput(podClient PodClient, artifactDir, namespace, podName string) error {
+	var validationErrors []error
+	list, err := podClient.Pods(namespace).List(meta.ListOptions{FieldSelector: fields.Set{"metadata.name": podName}.AsSelector().String()})
+	if err != nil {
+		return fmt.Errorf("could not list pod: %v", err)
+	}
+	pod := &list.Items[0]
+
+	if pod.Annotations["ci-operator.openshift.io/containers-logged-on-failure"] == "" {
```
smarterclayton: I think the case was made earlier that this could be "container-logs-as-artifacts" (don't have to deal with failure).  That's simpler, and getting the extra logs isn't that bad

pkg/steps/artifacts.go
```diff
@@ -505,3 +511,65 @@ func allPodContainerNamesWithArtifacts(pod map[string]interface{}) map[string]st
...
+			if _, exists := containersToOutput[status.Name]; exists {
+				file, err := os.Create(fmt.Sprintf("%s/%s.log", artifactDir, status.Name))
+				if err != nil {
+					validationErrors = append(validationErrors, fmt.Errorf("Cannot create file: %v", err))
+					continue
+				}
+				defer file.Close()
+
+				if s, err := podClient.Pods(namespace).GetLogs(podName, &coreapi.PodLogOptions{Container: status.Name}).Stream(); err == nil {
+					if _, err := io.Copy(file, s); err != nil {
```
smarterclayton: We should gzip the output

pkg/steps/artifacts.go
```diff
@@ -505,3 +511,65 @@ func allPodContainerNamesWithArtifacts(pod map[string]interface{}) map[string]st
...
+	if pod.Annotations["ci-operator.openshift.io/containers-logged-on-failure"] == "" {
+		return nil
+	}
+
+	if err := os.MkdirAll(artifactDir, 0750); err != nil {
+		return fmt.Errorf("unable to create directory %s: %v", artifactDir, err)
+	}
+
+	var containersToLog []string
+	if names := pod.Annotations["ci-operator.openshift.io/containers-logged-on-failure"]; len(names) > 0 {
```
smarterclayton: I would just make this be per pod: `ci-operator.openshift.io/save-container-logs`

droslean: So, we don't have to be selective? Just `ci-operator.openshift.io/save-container-logs: "true"` and save the output for all the containers of the pod?

droslean:
Deleting the namespace and `retest` solved the problem.

pkg/steps/artifacts.go
smarterclayton: yes

smarterclayton:
Have you tested this on a job directly (to verify the artifacts work)?

droslean:
Yes, I have tested it locally.

bbguimaraes:
There doesn't seem to be any remaining impediments.

/lgtm

PR #207, opened 2018-11-13
Add more profiles for openshift-ansible 4.0 testing

vrutkovs:
This adds separate configs for BYOR 4.0 on AWS and GCP


bbguimaraes:
@vrutkovs: you will need to change the validation code in `pkg/api/config.go` if you want to use them with `ci-operator`.

vrutkovs:
Yep, will do

PR #209, opened 2018-11-13
Always pull pipeline images

kargakis:
Avoid reusing cached, possibly invalid, images.

~~Fixes https://github.com/openshift/release/issues/2055~~


kargakis:
I just realized we need to set the policy in our templates. Still useful for ci-operator @stevekuznetsov ?

smarterclayton:
/hold

This should never be a thing.  If you are rebuilding a namespace and overriding pipeline tags you are "doing it wrong" (ci-operator is not intended to rebuild over things in a namespace)

kargakis:
> ci-operator is not intended to rebuild over things in a namespace

This is how ci-operator works

kargakis:
can be a jira card if it needs to change

PR #212, opened 2018-11-14
add gcp cluster profiles for logging

richm:
There are 3 new gcp cluster profiles for logging, one each
for docker log-driver journald, docker log-driver json-file,
and crio


richm:
@jcantrill fyi

PR #213, opened 2018-11-21
Unit tests for some more straightforward steps

petr-muller:
Before trying to unit test some of the larger, complicated steps that would actually need to mock some cluster state, I'm trying to cover some of the simpler ones. I have tried to extract some testing routines to a separate `testing.go` file and I'm trying to use that for testing all the steps. I needed to grok what `DeferredParams` in `template.go` does, so I wrote unit tests for it, discovering some suspicious parts in the process:

- Map() method does not perform the `os.LookupEnv(name)` like the other read-ish methods do, so `Map()` can return something not consistent with what `Get()` would return.
- `DeferredParams.links` is a slice (`[]api.StepLink)`, but it does not ever contain more than a single element. The only method that writes it is `Add()` and that only sets it to be a single-item slice. So either this is wrong or it could be just a straight `api.StepLink`, not a slice.

I'm mostly submitting this to get opinions on the `testing.go` stuff and how the actual tests look like when using that code.

### DIFF
--- a/pkg/steps/template_test.go
+++ b/pkg/steps/template_test.go
@@ -1,0 +1,221 @@
+package steps
+
+import (
+	"reflect"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/diff"
+
+	"github.com/openshift/ci-operator/pkg/api"
+)
+
+func TestDeferredParametersAllLinks(t *testing.T) {
+	var testCases = []struct {
+		purpose       string
+		dp            *DeferredParameters
+		expectedItems int
+	}{{
+		purpose: "AllLinks should return a slice with all links for all names",
+		dp: &DeferredParameters{
+			links: map[string][]api.StepLink{
+				"K1": {someStepLink("ONE"), someStepLink("TWO")},
+				"K2": {someStepLink("THREE")},
+			},
+		},
+		// only compare count of returned items because AllLinks can return links
+		// in any order and StepLink instances cannot be sorted unless the interface
+		// is extended to provide something by which to sort
+		expectedItems: 3,
+	}}
+
+	for _, tc := range testCases {
+		links := tc.dp.AllLinks()
+
+		if len(links) != tc.expectedItems {
+			t.Errorf("%s: %v.AllLinks() returned %v, expected %d items", tc.purpose, tc.dp, links, tc.expectedItems)
+		}
+	}
+}
+
+func TestDeferredParametersMap(t *testing.T) {
+	var testCases = []struct {
+		purpose  string
+		dp       *DeferredParameters
+		expected map[string]string
+	}{{
+		purpose: "values[N]=V, fns[N] is not set, so returned map does not have key 'N'",
+		dp: &DeferredParameters{
+			values: map[string]string{"K1": "V1"},
+			fns:    map[string]func() (string, error){},
+		},
+		expected: map[string]string{},
+	}, {
+		purpose: "fns[N] is set, values[N] is not, so returned map has key 'N' set to fns[N]()",
+		dp: &DeferredParameters{
+			values: map[string]string{},
+			fns:    map[string]func() (string, error){"K1": func() (string, error) { return "V1", nil }},
+		},
+		expected: map[string]string{"K1": "V1"},
+	}, {
+		purpose: "fns[N] is set, values[N] is set, so returned map has key 'N' set to values[N]",
+		dp: &DeferredParameters{
+			values: map[string]string{"K1": "V1"},
+			fns:    map[string]func() (string, error){"K1": func() (string, error) { return "F1", nil }},
+		},
+		expected: map[string]string{"K1": "V1"},
+	}, {
+		purpose: "returned map contains all names",
+		dp: &DeferredParameters{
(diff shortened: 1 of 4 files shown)


pkg/steps/template_test.go
```diff
@@ -0,0 +1,173 @@
...
+		expectedItems int
+	}{{
+		purpose: "sanity",
+		dp: &DeferredParameters{
+			links: map[string][]api.StepLink{
+				"K1": {someStepLink("ONE"), someStepLink("TWO")},
+				"K2": {someStepLink("THREE")},
+			},
+		},
+		// only compare count of returned items because AllLinks can
```
stevekuznetsov: Can sort for deterministic ordering

pkg/steps/template_test.go
```diff
@@ -0,0 +1,173 @@
...
+	for _, tc := range testCases {
+		links := tc.dp.AllLinks()
+
+		if len(links) != tc.expectedItems {
+			t.Errorf("%s: %v.AllLinks() returned %v, expected %d items", tc.purpose, tc.dp, links, tc.expectedItems)
+		}
+	}
+}
+
+func TestDeferredParametersMap(t *testing.T) {
```
stevekuznetsov: Not being super familiar with the underlying code, I'm having trouble understanding the test. If we had more than one case it would be useful here, also a description/name that provided setup --> end result causality.

Is this test case ensuring that a func doesn't override a previous value? Does it enforce that a previous value without a func is not exposed? Are there other cases? Can we break them up into separate cases?

pkg/steps/template_test.go
```diff
@@ -0,0 +1,173 @@
...
+			links:  map[string][]api.StepLink{"key": {someStepLink("old-link")}},
+		},
+		name:  "key",
+		link:  someStepLink("new-link"),
+		fn:    func() (string, error) { return "new", nil },
+		has:   true,
+		links: []api.StepLink{someStepLink("new-link")},
+	}}
+	for _, tc := range testCases {
+		if tc.name != "DO-NOT-ADD" {
```
stevekuznetsov: This smells of a bad test array setup?

pkg/steps/template_test.go
```diff
@@ -0,0 +1,173 @@
...
+			fns:    make(api.ParameterMap),
+			values: map[string]string{"key": "oldValue"},
+			links:  map[string][]api.StepLink{},
+		},
+		name:     "key",
+		setValue: "newValue",
+		getValue: "oldValue",
+		getError: nil,
+	}, {
+		purpose: "Existing key is not set if lazy evaluation func is set",
```
stevekuznetsov: Ok, these names rock :100:

pkg/steps/template_test.go
petr-muller: Not that simply, seems like I would need to implement some interface to make those classes sortable and I was lazy. I'll consider it (and maybe I'll even get rid of misusing the "real" StepLink structs and implement a simple one for testing).

pkg/steps/template_test.go
petr-muller: I'll try to look into this to make it more readable, at least with descriptions and/or names.

pkg/steps/template_test.go
petr-muller: Indeed :)

petr-muller:
Thanks for the feedback, I'll try to incorporate it. I was working on testing the `test.go` step in the meantime, which finally does some testing against fakes, and I did some changes, so I need to reconcile all this work into some sensible changes.

pkg/steps/template_test.go
petr-muller: IMO not worth the effort, `StepLink` is an interface that can do only `Same` and `Matches`, there's nothing to sort them by easily :/

pkg/steps/template_test.go
petr-muller: I've broken the testcase to smaller, more targeted ones. You were right I was probably trying to do too much in a single one.

pkg/steps/template_test.go
petr-muller: Replaced with explicit `callAdd` / `callSet` booleans.

petr-muller:
I've addressed the feedback and also incorporated some followup work, mainly the fact that I've separated the original `examineStep` which tested all `Step` methods into two smaller ones. The `examineStep` checks the "static" methods (not related to step execution) and `executeStep` that checks the `Done()` and `Run()` methods. The rationale for that is that is that especially `Run()` is likely to be checked in different scenarios for some steps (different cluster behavior, different inputs etc) and it is useless to check everything all the time.

Ready for re-review.

pkg/steps/template_test.go
```diff
@@ -0,0 +1,220 @@
+package steps
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/openshift/ci-operator/pkg/api"
+	"k8s.io/apimachinery/pkg/util/diff"
```
stevekuznetsov: nit: group imports, first stdlib, then third party, then ci-operator

petr-muller: Addressed.

PR #215, opened 2018-11-26
Style fixes to silence gofmt called by `make check`

petr-muller:


petr-muller:
Oh hell, it's a different go version :/ My Fedora has 1.11, we run 1.10 :-1: 

/close

stevekuznetsov:
We can bump to 1.11 if you'd like -- we will need a new `openshift/release` image for that

PR #216, opened 2018-11-26
Fix a possible deadlock in waitForCompletionOrTimeout

petr-muller:
There is a small window between first checking the Pod status on
L586/L590 and setting up the watcher on L594. If the Pod status changes
exactly in this window, we may end up waiting indefinitely on L604 for
an event that already happened.

Although this is very unlikely to happen in reality, I was seeing it
happen occasionally when unit-testing with fakes.


stevekuznetsov:
This pattern is used in some of the other flows as well -- can we change them as well in a follow-up? Builds are one off the top of my head.

/lgtm

petr-muller:
> This pattern is used in some of the other flows as well -- can we change them as well in a follow-up? Builds are one off the top of my head.

I'll try to look for the obvious occurrences.

PR #219, opened 2018-11-27
Unit tests for test.go step 

petr-muller:
These tests finally make use of the cluster mocking libraries. The mock
Pods.Create() method needed to be extended to create Pod with
Status.Phase=Pending, because the libraries do not put any defaults into
the fields and ci-operator code was failing on Phase="".

The `executeStep` method can now be provided a function that is executed
in a goroutine before the `Run()` method is executed, and which can be
used to simulate the expected fake cluster behavior.

This builds on https://github.com/openshift/ci-operator/pull/213

/cc @stevekuznetsov @bbguimaraes @droslean


pkg/steps/test_test.go
```diff
@@ -0,0 +1,160 @@
+package steps
```
stevekuznetsov: `test_test.go`

lol

pkg/steps/test_test.go
```diff
@@ -0,0 +1,160 @@
...
+			// Immediately set the Pod status to Succeeded, because
+			// that is what the step waits on
+			for {
+				event, ok := <-watcher.ResultChan()
+				if !ok {
+					break
+				}
+				if pod, ok := event.Object.(*v1.Pod); ok {
+					pod.Status.Phase = tc.podStatus
+					client.Pods(namespace).UpdateStatus(pod)
```
stevekuznetsov: Are we ignoring the returned error here?

pkg/steps/test_test.go
petr-muller: Serious question: should we rename that file to something like `container.go` or `pod.go`? I was considering that.

petr-muller:
Added some logging and error condition checks. Started to use the `t.Run()` pattern for better logging and earlier cleanups with `defer`. Fixed a test data race.

pkg/steps/test_test.go
stevekuznetsov: Yeah, you could do that for readability

pkg/steps/test_test.go
```diff
@@ -0,0 +1,179 @@
...
+			}
+
+			watcher, err := client.Pods(namespace).Watch(meta.ListOptions{})
+			if err != nil {
+				t.Errorf("Failed to create a watcher over pods in namespace")
+			}
+			defer watcher.Stop()
+
+			clusterBehavior := func() {
+				// Expect a single event (a Creation) to happen
```
stevekuznetsov: @smarterclayton is this appropriate?

pkg/steps/test_test.go
```diff
@@ -0,0 +1,179 @@
...
+				// Immediately set the Pod status to Succeeded, because
+				// that is what the step waits on
+				for {
+					event, ok := <-watcher.ResultChan()
+					if !ok {
+						t.Log("Fake cluster: watcher event closed, exiting()")
```
stevekuznetsov: Shouldn't this be an error?

pkg/steps/test_test.go
```diff
@@ -0,0 +1,179 @@
...
+						break
+					}
+					if pod, ok := event.Object.(*v1.Pod); ok {
+						t.Logf("Fake cluster: Received event on pod '%s': %s", pod.ObjectMeta.Name, event.Type)
+						t.Logf("Fake cluster: Updating pod '%s' status to '%s' and exiting", pod.ObjectMeta.Name, tc.podStatus)
+						// make a copy to avoid a race
+						newPod := *pod
```
stevekuznetsov: Use the `DeepCopy()` func instead

pkg/steps/test_test.go
```diff
@@ -0,0 +1,179 @@
...
+						}
+						break
+					}
+					t.Logf("Fake cluster: Received non-pod event: %v", event)
+				}
+			}
+
+			executeStep(t, ps, executionExpectation, clusterBehavior)
+
+			if pod, err := client.Pods(namespace).Get(ps.Name(), meta.GetOptions{}); !reflect.DeepEqual(expectedPod, pod) {
```
stevekuznetsov: When comparing API objects you need `semantic.Equality.DeepEqual()`

stevekuznetsov: (in some objects the order of a slice may not be guaranteed or semantically important for instance)

pkg/steps/test_test.go
petr-muller: Cool, did not know that one.

pkg/steps/test_test.go
petr-muller: Probably yes

pkg/steps/test_test.go
petr-muller: Addressed.

pkg/steps/test_test.go
petr-muller: Renamed to `pod.go`

petr-muller:
Addressed last round of review.

petr-muller:
@smarterclayton I don't plan to over-invest. We had a discussion within the team (I guess we should involve you) and we thought "each step has at least one positive test" is a reasonable goal to kick-off tests for ci-operator, so that we can start to nag ourselves to write more tests when we submit PRs for ci-operator (touching a piece of code? test it for it) which is uncomfortable to do when there are *none*. 

I agree these tests are kinda middle-level, but I still think they are useful - they end up verifying statements like "given a step config and an initial cluster state, after we execute the step, cluster is in a target state", together with the rest of the `Step` interface (the methods that the graph execution uses). To me, that matches to the "given input X, we should generate a tag with an image Y" description you describe. Anything more high-level is IMO hard to run against a mock cluster, because you would need to simulate more cluster behavior (submitted pods terminating, submitted builds terminating and producing outputs...) and that is probably the most fragile thing involved. Could you elaborate more about how you envision the high-level tests? What would be the input X and what kind of assertions would you perform on image Y?

I agree low-level tests would be helpful too, but that effort would be larger - there is simply too much low-level code to be covered. It could be prioritized, but I do not have sufficient knowledge about what would be the pieces most important to cover.

smarterclayton:
Re: high level, don’t mock a cluster, run a controlled config that exercises a set of steps against the cluster.  We call this an e2e (end to end test) and even for this repo two or three that gated merge would prevent total regressions of our ci systems.  That’s similar to why we have e2e-aws gating *all* repos - you can’t regress the core accidentally.

Re: low level, focus on areas that have had high bigs.  Ie waitForConpletion.  Ensure the code + usage is testable, and that you can simulate realistic scenarios (i.e. a set of pod changes coming down a watch).

The current tests aren’t wrong.  Just don’t be afraid to test in different ways earlier rather than later.  Having 20% coverage of three layers is better than 60% coverage of one “layer”.

stevekuznetsov:
Shit. @petr-muller let's rebase and merge these. @bbguimaraes @droslean did you review?

petr-muller:
No big deal, these are easy to rebase

PR #222, opened 2018-11-29
WIP:  Tests for the InputImageTag step 

petr-muller:
Test prepares an ImageStreamTag in one namespace with the necessary fields, executes a matching InputImageTagStep and verifies another ImageStreamTag was created in a target namespace.

The test does not need any fake cluster behavior.

Builds on https://github.com/openshift/ci-operator/pull/219 - until that gets merged see just the last commit here.

/cc @stevekuznetsov @droslean @bbguimaraes


petr-muller:
Rebased after the parent unit-test PR was merged, so this is no longer a WIP and can be reviewed quite easily.

stevekuznetsov:
This one is straightforward
/lgtm

PR #225, opened 2018-12-03
Move the real clonerefs file, not the symlink

stevekuznetsov:

/assign @smarterclayton


petr-muller:
> @petr-muller what is going on in that test?

I have no idea :D

stevekuznetsov:
It was fine, I was dumb

PR #226, opened 2018-12-04
Tests for the OutputImageTag step

petr-muller:
The `OutputImageTag` step is the first one tested that actually produces something in its `Provides()` method, which required little changes in how the method is tested. Previously, the test helper attempted to compare returned `func` instances which did not make sense. These functions are pure though, which allows them to be tested by calling them and comparing their return values instead.

Builds on #222


petr-muller:
Parent PR was merged, this one is ready to be reviewed.

stevekuznetsov:
@bbguimaraes you want to get this one?

bbguimaraes:
Oops, I ended up neglecting this last week… It looks simple enough. I like how easy to follow and clear these tests are.

/lgtm

PR #229, opened 2018-12-11
Unit tests: RPMImageInjectionStep 

petr-muller:
This step is the first from the unit-tested to use `Route`s and `Build`s, so
fake clientsets needed to be added for these resources. Testing the step
is straightforward and uses the established setup-examine-execute
pattern. This step depends on `Build`s having correct behavior (not just
`Build` resource having correct status at the end, but also creating the
Output, in this case, an `ImageStreamTag`), so fake cluster behavior is
more complicated. This behavior is likely to be useful for testing more
steps, so it was extracted into a separate method in testing.go.

Builds on https://github.com/openshift/ci-operator/pull/226

### DIFF
--- a/pkg/steps/rpm_injection_test.go
+++ b/pkg/steps/rpm_injection_test.go
@@ -1,0 +1,87 @@
+package steps
+
+import (
+	"testing"
+
+	buildv1 "github.com/openshift/api/build/v1"
+	routev1 "github.com/openshift/api/route/v1"
+
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	fakerest "k8s.io/client-go/rest/fake"
+
+	"github.com/openshift/ci-operator/pkg/api"
+)
+
+func TestRPMImageInjectionStep(t *testing.T) {
+	config := api.RPMImageInjectionStepConfiguration{
+		From: "configFrom",
+		To:   "configTo",
+	}
+	resources := api.ResourceConfiguration{}
+
+	fakecs := createTestingClient(t)
+
+	istclient := fakecs.imagecs.ImageV1()
+	routeclient := fakecs.routecs.RouteV1()
+	buildsgetter := fakecs.buildcs.BuildV1()
+	restclient := &fakerest.RESTClient{}
+	buildclient := NewBuildClient(buildsgetter, restclient)
+
+	jobspec := &api.JobSpec{Namespace: "jobspecNamespace"}
+
+	rpmiis := RPMImageInjectionStep(config, resources, buildclient, routeclient, istclient, jobspec)
+
+	specification := stepExpectation{
+		name: "configTo",
+		requires: []api.StepLink{
+			api.InternalImageLink("configFrom"),
+			api.RPMRepoLink(),
+		},
+		creates:  []api.StepLink{api.InternalImageLink("configTo")},
+		provides: providesExpectation{params: nil, link: nil},
+		inputs:   inputsExpectation{values: nil, err: false},
+	}
+
+	examineStep(t, rpmiis, specification)
+
+	rpmroute := &routev1.Route{
+		ObjectMeta: meta.ObjectMeta{Name: RPMRepoName},
+	}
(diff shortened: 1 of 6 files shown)


pkg/steps/testing.go
```diff
@@ -7,25 +7,45 @@ import (
...
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	fakecorev1 "k8s.io/client-go/kubernetes/typed/core/v1/fake"
 
+	"k8s.io/apimachinery/pkg/util/diff"
+	"k8s.io/client-go/kubernetes/fake"
+
 	"github.com/openshift/ci-operator/pkg/api"
 )
 
+func createTestingClient(t *testing.T) *ciopTestingClient {
```
droslean: Why here we need to pass a `*testing.T`?

pkg/steps/testing.go
```diff
@@ -110,6 +126,32 @@ func someStepLink(as string) api.StepLink {
 	})
 }
 
+func fakeSuccessfulBuild(build *apibuildv1.Build, buildclient buildv1.BuildV1Interface, istclient imagev1.ImageV1Interface, t *testing.T) {
```
droslean: is `Fake.PrependWatchReactor()` can do that?

pkg/steps/rpm_injection_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	routeclient.Routes(jobspec.Namespace).Create(rpmroute)
+
+	watcher, err := buildclient.Builds("jobspecNamespace").Watch(meta.ListOptions{})
+	if err != nil {
+		t.Errorf("Failed to create a watcher over Builds in namespace 'jobspecNamespace'")
+	}
+	defer watcher.Stop()
+
+	clusterBehavior := func() {
```
droslean: If this can be replaced with the following code, we don't need for the `fakeSuccessfulBuild()`.
```go
	clusterBehavior := func() {
		fakecs.buildcs.PrependWatchReactor("builds", func(action clientgotesting.Action) (bool, watch.Interface, error) {
			watcher.Stop()
			ret := watch.NewFakeWithChanSize(1, false)
			for {
				event, ok := <-watcher.ResultChan()
				if !ok {
					t.Error("Fake cluster: watcher event closed, exiting")
					break
				}
				if build, ok := event.Object.(*buildv1.Build); ok {
					t.Logf("Fake cluster: Received event on Build '%s': %s", build.ObjectMeta.Name, event.Type)
					t.Logf("Fake cluster: Updating build '%s' status to '%s' and exiting", build.ObjectMeta.Name, buildv1.BuildPhaseComplete)

					if build.Spec.Output.To.Kind == "ImageStreamTag" {
						name := build.Spec.Output.To.Name
						namespace := build.Spec.Output.To.Namespace
						t.Logf("Fake cluster: Build output is ImageStreamTag '%s' in namespace '%s': creating a fake ImageStreamTag with a fake Image", name, namespace)
						istag := &apiimagev1.ImageStreamTag{
							ObjectMeta: meta.ObjectMeta{
								Name:      name,
								Namespace: namespace,
							},
							Image: apiimagev1.Image{ObjectMeta: meta.ObjectMeta{Name: "DoesThisMatter"}},
						}
						if _, err := istclient.ImageStreamTags(namespace).Create(istag); err != nil {
							t.Errorf("Fake cluster: Failed to treate ImageStreamTag to fake Build output: %v", err)
						}
					}
					b := build.DeepCopy()
					b.Status.Phase = buildv1.BuildPhaseComplete
					ret.Modify(b)
					break
				}
				t.Logf("Fake cluster: Received non-build event: %v", event)
			}
			return true, ret, nil
		})
	}
```

either way, you understand my point.

stevekuznetsov:
@petr-muller what's left here?

petr-muller:
> @petr-muller what's left here?

Addressing Nikos' comments. I haven't really touched this in a while.

petr-muller:
Let's close this, I can always reopen when I actually have time to finish this.
/close

PR #230, opened 2018-12-14
If artifacts are requested, write metadata.json as gubernator expects

smarterclayton:
metadata.json is folded into finished.json by the prow entrypoint util,
and the ci-operator can write the fields it expects as well as our own
custom `work-namespace` variable to feed into gubernator and testgrid.

@stevekuznetsov

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -402,6 +402,10 @@ func (o *options) Run() error {
 			return fmt.Errorf("could not resolve inputs: %v", err)
 		}
 
+		if err := o.writeMetadataJSON(); err != nil {
+			return fmt.Errorf("unable to write metadata.json for build: %v", err)
+		}
+
 		// convert the full graph into the subset we must run
 		nodes, err := api.BuildPartialGraph(buildSteps, o.targets.values)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/ci-operator/main.go
```diff
@@ -690,6 +694,64 @@ func (o *options) initializeNamespace() error {
...
+}
+
+func (o *options) writeMetadataJSON() error {
+	if len(o.artifactDir) == 0 {
+		return nil
+	}
+
+	m := prowResultMetadata{}
+
+	specs := []string{fmt.Sprintf("%s:%s", o.jobSpec.Refs.BaseRef, o.jobSpec.Refs.BaseSHA)}
```
stevekuznetsov: There's a util func on Refs you can import to do this, or at least copy-paste it.

cmd/ci-operator/main.go
```diff
@@ -690,6 +694,64 @@ func (o *options) initializeNamespace() error {
...
+	for _, pull := range o.jobSpec.Refs.Pulls {
+		specs = append(specs, fmt.Sprintf("%d:%s", pull.Number, pull.SHA))
+	}
+	m.Repo = fmt.Sprintf("%s/%s", o.jobSpec.Refs.Org, o.jobSpec.Refs.Repo)
+	m.Repos = map[string]string{m.Repo: strings.Join(specs, ",")}
+
+	m.Pod = o.jobSpec.ProwJobID
+	m.WorkNamespace = o.namespace
+
+	data, err := json.MarshalIndent(m, "", "  ")
```
stevekuznetsov: Does the formatting really matter?

smarterclayton: Makes it much easier to read inside GCS when you're trying to debug stuff (the viewer shows it formatted like https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/71936/pull-kubernetes-e2e-gce/62557/finished.json instead of super compact and hard to read)

cmd/ci-operator/main.go
smarterclayton: ok

smarterclayton: We already had it.  Switched

PR #232, opened 2018-12-19
Usability improvements to validation for iteratively building a config.yaml

smarterclayton:
Surfaced a few improvements while debugging an issue in ci-operator around
message fallback output

* Better handling of validation and reporting to user
* Better messages for some error cases that guide user to next step
* Trim some whitespace from pod execution output
* Don't panic when the config file is empty


stevekuznetsov:
I like it

/lgtm

PR #245, opened 2019-01-21
Expose template's pod through 8080 & 8443 ports.

### LINKED ISSUE #244: Allow exposing the test pod using a Route
`openshift-acme` e2e tests need to be exposed to internet so Let's Encrypt can reach back and validate acme challenges. Have tried doing this whit `ngrok` before but they've introduced 20 req/s rate limit, then serveo.net but that strips the `.well-known` paths. Given the test (created) cluster is not exposed to the internet I can forward request from the e2e pod running in the api.ci cluster but I need that pod to be exposed by a Route. Then it's just a matter of `oc port-forward -n default dc/router 5000:80 &`. This could be helpful e.g. for integration tests, not just e2e.

I propose to extend the `ci-operator` config so every "test" has a `expose_domain` option leading to creation of a Route with that name. If the domain is empty the defaults apply, same for Routes. (And svc targeting the pod.). A think we can just settle for port 5000 and avoid additional config there.

/cc @droslean

droslean:
closes #244 

Creates a service and two routes for each template.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -24,6 +24,7 @@ import (
 	rbacapi "k8s.io/api/rbac/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	rbacclientset "k8s.io/client-go/kubernetes/typed/rbac/v1"
 	"k8s.io/client-go/rest"
@@ -35,9 +36,11 @@ import (
 
 	imageapi "github.com/openshift/api/image/v1"
 	projectapi "github.com/openshift/api/project/v1"
+	routev1 "github.com/openshift/api/route/v1"
 	templateapi "github.com/openshift/api/template/v1"
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	"github.com/openshift/client-go/project/clientset/versioned"
+	routeclientset "github.com/openshift/client-go/route/clientset/versioned"
 	templatescheme "github.com/openshift/client-go/template/clientset/versioned/scheme"
 
 	"github.com/openshift/ci-operator/pkg/api"
@@ -200,6 +203,7 @@ type options struct {
 	clusterConfig *rest.Config
 
 	givePrAuthorAccessToNamespace bool
+	exposeTemplatePod             bool
 	impersonateUser               string
 }
 
@@ -241,6 +245,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.StringVar(&opt.gitRef, "git-ref", "", "Populate the job spec from this local Git reference. If JOB_SPEC is set, the refs field will be overwritten.")
 	flag.BoolVar(&opt.givePrAuthorAccessToNamespace, "give-pr-author-access-to-namespace", false, "Give view access to the temporarily created namespace to the PR author.")
 	flag.StringVar(&opt.impersonateUser, "as", "", "Username to impersonate")
+	flag.BoolVar(&opt.exposeTemplatePod, "expose-template-pod", false, "Expose the template's pod by creating a service and two routes for ports 8080 and 8443.")
 
 	return opt
 }
@@ -357,6 +362,8 @@ func (o *options) Complete() error {
 			template.Name = filepath.Base(path)
 			template.Name = strings.TrimSuffix(template.Name, filepath.Ext(template.Name))
 		}
+
+		template.ObjectLabels = map[string]string{"ci-operator.openshift.io/template-name": template.Name}
 		o.templates = append(o.templates, template)
 	}
 
@@ -599,6 +606,17 @@ func (o *options) initializeNamespace() error {
 	client, err := coreclientset.NewForConfig(o.clusterConfig)
 	if err != nil {
 		return fmt.Errorf("could not get core client for cluster config: %v", err)
+	}
+
+	routeClient, err := routeclientset.NewForConfig(o.clusterConfig)
+	if err != nil {
+		return fmt.Errorf("could not get route client for cluster config: %v", err)
+	}
+
+	if o.exposeTemplatePod {
+		if err := createTemplateServiceAndRoutes(client, routeClient, o.templates, o.namespace); err != nil {
+			log.Printf("warning: %v", err)
+		}
 	}
 
 	updates := map[string]string{}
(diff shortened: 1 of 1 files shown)


cmd/ci-operator/main.go
```diff
@@ -22,10 +22,13 @@ import (
 
 	coreapi "k8s.io/api/core/v1"
 	rbacapi "k8s.io/api/rbac/v1"
+
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	rbacclientset "k8s.io/client-go/kubernetes/typed/rbac/v1"
+
```
tnozicka: the usual openshift convention is to group k8s.io, openshift, standard and other imports

cmd/ci-operator/main.go
```diff
@@ -241,6 +245,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.StringVar(&opt.gitRef, "git-ref", "", "Populate the job spec from this local Git reference. If JOB_SPEC is set, the refs field will be overwritten.")
 	flag.BoolVar(&opt.givePrAuthorAccessToNamespace, "give-pr-author-access-to-namespace", false, "Give view access to the temporarily created namespace to the PR author.")
 	flag.StringVar(&opt.impersonateUser, "as", "", "Username to impersonate")
+	flag.BoolVar(&opt.exposeTemplatePod, "expose-template-pod", false, "Expose the template's pod by creating a service and two routes for ports 8080 and 8443.")
```
tnozicka: routes are actually for ports 80 and 443, but end up being redirected as 8080 and 8443 to the pod itself

cmd/ci-operator/main.go
```diff
@@ -357,6 +362,8 @@ func (o *options) Complete() error {
 			template.Name = filepath.Base(path)
 			template.Name = strings.TrimSuffix(template.Name, filepath.Ext(template.Name))
 		}
+
+		template.ObjectLabels = map[string]string{"ci-operator.openshift.io/template-name": template.Name}
```
tnozicka: make the key a const

cmd/ci-operator/main.go
```diff
@@ -601,6 +608,17 @@ func (o *options) initializeNamespace() error {
...
 	}
 
+	routeClient, err := routeclientset.NewForConfig(o.clusterConfig)
+	if err != nil {
+		return fmt.Errorf("could not get route client for cluster config: %v", err)
+	}
+
+	if o.exposeTemplatePod {
+		if err := createTemplateServiceAndRoutes(client, routeClient, o.templates, o.namespace); err != nil {
+			log.Printf("warning: %v", err)
```
tnozicka: does it ignore the error?

cmd/ci-operator/main.go
```diff
@@ -933,3 +951,70 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+						Protocol:   "TCP",
+						Port:       80,
+						TargetPort: intstr.FromInt(8080),
+					},
+					{
+						Name:       "https",
+						Protocol:   "TCP",
+						Port:       443,
+						TargetPort: intstr.FromInt(8443),
+					}}}}
```
tnozicka: nit: I'd put new lines in between, easier to match blocks then

cmd/ci-operator/main.go
```diff
@@ -933,3 +951,70 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+
+		if _, err := client.Services(namespace).Create(s); err != nil {
+			return fmt.Errorf("could not create service %s: %v", template.Name, err)
+		}
+
+		// Create 8080 Route
```
tnozicka: the route is actually 80 (applies globally)

cmd/ci-operator/main.go
```diff
@@ -933,3 +951,70 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+		if _, err := routeClient.Route().Routes(namespace).Create(&routev1.Route{
+			ObjectMeta: meta.ObjectMeta{
+				Name: fmt.Sprintf("%s-8080-route", template.Name),
+			},
+			Spec: routev1.RouteSpec{
+				TLS: &routev1.TLSConfig{
+					Termination: "passthrough",
```
tnozicka: const - https://github.com/openshift/origin/blob/cf3c5d31570d6346a1239d5e751a5b4bdaa15bf7/pkg/route/apis/route/types.go#L181

cmd/ci-operator/main.go
```diff
@@ -933,3 +951,70 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+				},
+				To: routev1.RouteTargetReference{Name: template.Name},
+				Port: &routev1.RoutePort{
+					TargetPort: intstr.FromInt(8080),
```
tnozicka: i'd think this should be 80 and the redirection is done on the service itself

cmd/ci-operator/main.go
```diff
@@ -933,3 +951,70 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+				},
+			},
+		}); err != nil && !kerrors.IsAlreadyExists(err) {
```
tnozicka: you should fail if it exists as it is creating a new namespace. in general you can't be sure about the configuration of an existing route

cmd/ci-operator/main.go
```diff
@@ -933,3 +951,70 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+		}
+
+		// Create 8443 Route
+		if _, err := routeClient.Route().Routes(namespace).Create(&routev1.Route{
+			ObjectMeta: meta.ObjectMeta{
+				Name: fmt.Sprintf("%s-8443-route", template.Name),
+			},
+			Spec: routev1.RouteSpec{
+				TLS: &routev1.TLSConfig{
+					Termination: "passthrough",
```
tnozicka: same here

cmd/ci-operator/main.go
```diff
@@ -933,3 +951,70 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+				},
+				To: routev1.RouteTargetReference{Name: template.Name},
+				Port: &routev1.RoutePort{
+					TargetPort: intstr.FromInt(8443),
+				},
+			},
+		}); err != nil && !kerrors.IsAlreadyExists(err) {
```
tnozicka: same here

tnozicka:
left few comments, but I am not picky - whatever exposes the pod, makes me happy.

@droslean thanks for working on that!

cmd/ci-operator/main.go
```diff
@@ -933,3 +951,70 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+				Port: &routev1.RoutePort{
+					TargetPort: intstr.FromInt(8080),
+				},
+			},
+		}); err != nil && !kerrors.IsAlreadyExists(err) {
+			return fmt.Errorf("could not create route %s-8080-route: %v", template.Name, err)
+		}
+
+		// Create 8443 Route
+		if _, err := routeClient.Route().Routes(namespace).Create(&routev1.Route{
```
tnozicka: this route will have a different host/domain, unless you set the `host` to the host substituted into the first route here manually to match

cmd/ci-operator/main.go
```diff
@@ -933,3 +951,70 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+						Port:       443,
+						TargetPort: intstr.FromInt(8443),
+					}}}}
+
+		if _, err := client.Services(namespace).Create(s); err != nil {
+			return fmt.Errorf("could not create service %s: %v", template.Name, err)
+		}
+
+		// Create 8080 Route
+		if _, err := routeClient.Route().Routes(namespace).Create(&routev1.Route{
```
tnozicka: don't throw away the runtime.object - host is substituted on admission so it will be already set in the returned object.

PR #247, opened 2019-01-25
Allow pods to declare that they want one junit per container, instead of pod

smarterclayton:
Setting the annotation `ci-operator.openshift.io/container-sub-tests` to a
list of comma delimited container names will replace the overall junit
result reported for that pod or template with one junit result per test.

This allows a pod that has both setup and test containers to get per task
junits which aids debugging.

Also simplify the names of the steps as printed in junit to make them
easier to scan.

```
<testsuites>
  <testsuite name="operator" tests="1" skipped="0" failures="1" time="5.068806427">
    <testcase name="Run template template - test container test" time="0">
      <failure message="">nameserver 10.142.0.22&#xA;search clayton-test-1.svc.cluster.local svc.cluster.local cluster.local c.openshift-ci-infra.internal google.internal&#xA;options ndots:5&#xA;</failure>
    </testcase>
  </testsuite>
</testsuites>
```

for the example in the dir (time is zero because kube accuracy is +/- 1 second)

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -219,9 +219,10 @@ type ArtifactWorker struct {
 
 	podsToDownload chan string
 
-	lock      sync.Mutex
-	remaining podContainersMap
-	required  podContainersMap
+	lock         sync.Mutex
+	remaining    podContainersMap
+	required     podContainersMap
+	hasArtifacts map[string]struct{}
 }
 
 func NewArtifactWorker(podClient PodClient, artifactDir, namespace string) *ArtifactWorker {
@@ -231,8 +232,9 @@ func NewArtifactWorker(podClient PodClient, artifactDir, namespace string) *Arti
 		namespace: namespace,
 		dir:       artifactDir,
 
-		remaining: make(podContainersMap),
-		required:  make(podContainersMap),
+		remaining:    make(podContainersMap),
+		required:     make(podContainersMap),
+		hasArtifacts: make(map[string]struct{}),
 
 		podsToDownload: make(chan string, 4),
 	}
@@ -242,7 +244,8 @@ func NewArtifactWorker(podClient PodClient, artifactDir, namespace string) *Arti
 
 func (w *ArtifactWorker) run() {
 	for podName := range w.podsToDownload {
-		if err := w.downloadArtifacts(podName); err != nil {
+		_, hasArtifacts := w.hasArtifacts[podName]
+		if err := w.downloadArtifacts(podName, hasArtifacts); err != nil {
 			log.Printf("error: %v", err)
 		}
 		// indicate we are done with this pod by removing the map entry
@@ -252,7 +255,19 @@ func (w *ArtifactWorker) run() {
 	}
 }
 
-func (w *ArtifactWorker) downloadArtifacts(podName string) error {
+func (w *ArtifactWorker) downloadArtifacts(podName string, hasArtifacts bool) error {
+	if err := os.MkdirAll(w.dir, 0750); err != nil {
+		return fmt.Errorf("unable to create artifact directory %s: %v", w.dir, err)
+	}
+	if err := gatherContainerLogsOutput(w.podClient, filepath.Join(w.dir, "container-logs"), w.namespace, podName); err != nil {
+		log.Printf("error: unable to gather container logs: %v", err)
+	}
+
+	// only pods with an artifacts container should be gathered
+	if !hasArtifacts {
+		return nil
+	}
+
 	defer func() {
 		// signal to artifacts container to gracefully shut don
 		err := removeFile(w.podClient, w.namespace, podName, "artifacts", []string{"/tmp/done"})
@@ -271,12 +286,6 @@ func (w *ArtifactWorker) downloadArtifacts(podName string) error {
 		// give up, expect another process to clean up the pods
 	}()
 
-	if err := os.MkdirAll(w.dir, 0750); err != nil {
-		return fmt.Errorf("unable to create artifact directory %s: %v", w.dir, err)
-	}
-	if err := gatherContainerLogsOutput(w.podClient, filepath.Join(w.dir, "container-logs"), w.namespace, podName); err != nil {
-		log.Printf("error: unable to gather container logs: %v", err)
-	}
 	if err := copyArtifacts(w.podClient, w.dir, w.namespace, podName, "artifacts", []string{"/tmp/artifacts"}); err != nil {
 		return fmt.Errorf("unable to retrieve artifacts from pod %s: %v", podName, err)
 	}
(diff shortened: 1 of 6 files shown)


smarterclayton:
This will give us accurate test grid timings of tests only, teardown only, and install only (so we can look for regressions in those components).  Will also make initial triage of failures easier.

smarterclayton:
@bbguimaraes @petr-muller fyi

smarterclayton:
Also @stevekuznetsov

pkg/steps/artifacts.go
```diff
@@ -24,6 +24,7 @@ import (
 	"k8s.io/client-go/tools/remotecommand"
 
 	templateapi "github.com/openshift/api/template/v1"
+	"github.com/openshift/ci-operator/pkg/junit"
```
stevekuznetsov: nit: can we get this import stanza broken up correctly

pkg/steps/artifacts.go
```diff
@@ -51,6 +52,72 @@ func (nopNotifier) Complete(_ string)               {}
 func (nopNotifier) Done(_ string) bool              { return true }
 func (nopNotifier) Cancel()                         {}
 
+// TestCaseNotifier allows a caller to generate per container unit tests
```
stevekuznetsov: `s/unit tests/jUnit records/`?

pkg/steps/artifacts.go
```diff
@@ -51,6 +52,72 @@ func (nopNotifier) Complete(_ string)               {}
...
+// that provide better granularity for debugging problems. It intercepts
+// notifications and remembers the last pod retrieved.
+type TestCaseNotifier struct {
+	nested  ContainerNotifier
+	lastPod *coreapi.Pod
+}
+
+func NewTestCaseNotifier(nested ContainerNotifier) *TestCaseNotifier {
+	return &TestCaseNotifier{nested: nested}
```
stevekuznetsov: Why not just export the field?

pkg/steps/artifacts.go
```diff
@@ -51,6 +52,72 @@ func (nopNotifier) Complete(_ string)               {}
...
+// SubTests returns one junit test for each terminated container with a name
+// in the annotation 'ci-operator.openshift.io/container-sub-tests' in the pod.
+// Invoking SubTests clears the last pod, so subsequent calls will return no
+// tests unless Notify() has been called in the meantime.
+func (n *TestCaseNotifier) SubTests(prefix string) []*junit.TestCase {
+	if n.lastPod == nil {
+		return nil
+	}
+	pod := n.lastPod
+	n.lastPod = nil
```
stevekuznetsov: Is this threadsafe?

pkg/steps/artifacts.go
```diff
@@ -51,6 +52,72 @@ func (nopNotifier) Complete(_ string)               {}
...
+
+	names := strings.Split(pod.Annotations["ci-operator.openshift.io/container-sub-tests"], ",")
```
stevekuznetsov: Let's get us a constant

pkg/steps/artifacts.go
```diff
@@ -51,6 +52,72 @@ func (nopNotifier) Complete(_ string)               {}
...
+	if len(names) == 1 && names[0] == "" {
+		return nil
+	}
+	var tests []*junit.TestCase
+	for _, status := range pod.Status.ContainerStatuses {
```
stevekuznetsov: No check for `""` in the comma-delimited list?

pkg/steps/artifacts.go
```diff
@@ -51,6 +52,72 @@ func (nopNotifier) Complete(_ string)               {}
...
+		if stringInSlice(names, status.Name) {
```
stevekuznetsov:
```go
if sets.NewString(names...).Has(status.Name) {
```

?

pkg/steps/artifacts.go
```diff
@@ -219,9 +286,10 @@ type ArtifactWorker struct {
 
 	podsToDownload chan string
 
-	lock      sync.Mutex
-	remaining podContainersMap
-	required  podContainersMap
+	lock         sync.Mutex
+	remaining    podContainersMap
+	required     podContainersMap
+	hasArtifacts map[string]struct{}
```
stevekuznetsov:
```
sets.String()
```

pkg/steps/artifacts.go
```diff
@@ -51,6 +52,72 @@ func (nopNotifier) Complete(_ string)               {}
...
+}
+
+func (n *TestCaseNotifier) Complete(podName string)  { n.nested.Complete(podName) }
+func (n *TestCaseNotifier) Done(podName string) bool { return n.nested.Done(podName) }
+func (n *TestCaseNotifier) Cancel()                  { n.nested.Cancel() }
+
+// SubTests returns one junit test for each terminated container with a name
+// in the annotation 'ci-operator.openshift.io/container-sub-tests' in the pod.
+// Invoking SubTests clears the last pod, so subsequent calls will return no
+// tests unless Notify() has been called in the meantime.
```
stevekuznetsov: Not sure I fully understand the point of this caching mechanism

test/template.yaml
```diff
@@ -6,6 +6,8 @@ objects:
   apiVersion: v1
   metadata:
     name: test
+    annotations:
+      ci-operator.openshift.io/container-sub-tests: test
```
stevekuznetsov: Is it possible for us to write a test that would show us this behavior actually happening?

pkg/steps/artifacts.go
smarterclayton: Just to be consistent with the others

pkg/steps/artifacts.go
smarterclayton: Today we only use it in single threaded contexts, there's some affordances built in to tolerate multithreaded notification (basically setup single thread, single threaded notify, parallel upload but doesn't read these fields, then single threaded junit extraction).   I can add comments to which methods are expected to be thread safe.

pkg/steps/artifacts.go
smarterclayton: I can, it couldn't cause a match (since we wouldn't find a container with "")

pkg/steps/artifacts.go
smarterclayton: We use the same notifier for multiple pods in a for loop in template processing.  So if you have template with pods A and B, we pass the same notifier to A that wee do to B.  That's also how artifacts are gathered right now (we wait until A passes, then check B).

test/template.yaml
smarterclayton: Probably.

pkg/steps/artifacts.go
stevekuznetsov: I see

smarterclayton:
Comments addressed

stevekuznetsov:
/lgtm
/hold
@petr-muller want to take a look?

petr-muller:
I'll look today

PR #250, opened 2019-01-30
Report only junit results for the main test container

smarterclayton:
Makes the output a lot cleaner


stevekuznetsov:
As opposed to all the `pod-utilities`?

/retest
/lgtm

smarterclayton:
In this case this is the pod we launch in the target namespace, but it avoids looking at "artifacts" which doesn't run a process, so the output is *only* the log output of the container under test which is much cleaner.

smarterclayton:
fixed test, reapplying label.

PR #253, opened 2019-01-30
test step secret inject

mjudeikis:
This will allow to inject secrets into test step pods. So we can test cloud interaction from test pods without the need to run complex template.

This is needed for openshift on azure work, so we could test cloud api's behaviour in test steps and keep all the logic inside openshift-azure repo, instead of release repo. 

/cc @jim-minter @stevekuznetsov

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -65,6 +65,9 @@ tests:
   commands: ''
   container:
     from: ''
+  secret:
+    name: ''
+    mount_path: ''
   openshift_ansible:
     cluster_profile: ''
   openshift_ansible_src:
@@ -312,6 +315,18 @@ of the images in the pipeline.
 
 ## `tests.container.from`
 `from` is the pipeline image tag that this test will be run on.
+
+# `tests.secret` 
+
+`Secret` field enables users to mount a secret inside test container.
+It is users responsibility to make sure secret is available in the temporary namespace.
+This can be done by providing flag `--secret-dir` to ci-operator in prow configuration.
+
+## `tests.secret.name`
+`secret.name` is the name of the secret to be mounted inside a test container.
+
+## `tests.secret.path`
+`secret.path` is the path at which to mount the secret. Optional, defaults to `/usr/test-secret`
 
 ## `tests.openshift_ansible`
 `openshift_ansible` is a test that provisions a cluster using openshift-ansible
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"path/filepath"
 	"regexp"
 	"strings"
 
@@ -104,6 +105,20 @@ func validateTestStepConfiguration(fieldRoot string, input []TestStepConfigurati
 
 		if len(test.Commands) == 0 {
 			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].commands: is required", fieldRoot, num))
+		}
+
+		if len(test.Secret.Name) > 0 {
+			// TODO: Move to upstream validation when vendoring is fixed
+			// currently checking against DNS RFC 1123 regexp
+			if ok := regexp.MustCompile("^[a-z0-9]([-a-z0-9]*[a-z0-9])?$").MatchString(test.Secret.Name); !ok {
+				validationErrors = append(validationErrors, fmt.Errorf("%s[%d].name: '%s' secret name is not valid value, should be [a-z0-9]([-a-z0-9]*[a-z0-9]", fieldRoot, num, test.Secret.Name))
+			}
+			// validate path only if name is passed
+			if test.Secret.MountPath != "" {
+				if ok := filepath.IsAbs(test.Secret.MountPath); !ok {
+					validationErrors = append(validationErrors, fmt.Errorf("%s[%d].path: '%s' secret mount path is not valid value, should be ^((\\/*)\\w+)+", fieldRoot, num, test.Secret.MountPath))
+				}
+			}
 		}
 
 		validationErrors = append(validationErrors, validateTestConfigurationType(fmt.Sprintf("%s[%d]", fieldRoot, num), test, release)...)
--- a/pkg/api/config_test.go
+++ b/pkg/api/config_test.go
@@ -155,6 +155,78 @@ func TestValidateTests(t *testing.T) {
 			release:       &ReleaseTagConfiguration{Name: "origin-v3.11"},
 			expectedValid: true,
 		},
+		{
+			id: "invalid secret mountPath",
+			tests: []TestStepConfiguration{
+				{
+					As: "test",
+					OpenshiftAnsibleClusterTestConfiguration: &OpenshiftAnsibleClusterTestConfiguration{},
+					Secret: Secret{
+						Name:      "secret",
+						MountPath: "/path/to/secret:exec",
+					},
+				},
+			},
+			expectedValid: false,
+		},
+		{
+			id: "invalid secret name",
+			tests: []TestStepConfiguration{
+				{
+					As: "test",
+					OpenshiftAnsibleClusterTestConfiguration: &OpenshiftAnsibleClusterTestConfiguration{},
+					Secret: Secret{
+						Name:      "secret_test",
+						MountPath: "/path/to/secret:exec",
+					},
+				},
+			},
+			expectedValid: false,
+		},
+		{
+			id: "valid secret",
+			tests: []TestStepConfiguration{
+				{
+					As:                         "unit",
+					Commands:                   "commands",
+					ContainerTestConfiguration: &ContainerTestConfiguration{From: "ignored"},
+					Secret: Secret{
+						Name: "secret",
+					},
+				},
+			},
+			expectedValid: true,
+		},
+		{
+			id: "valid secret with path",
+			tests: []TestStepConfiguration{
+				{
+					As:                         "unit",
+					Commands:                   "commands",
+					ContainerTestConfiguration: &ContainerTestConfiguration{From: "ignored"},
+					Secret: Secret{
+						Name:      "secret",
+						MountPath: "/path/to/secret",
+					},
+				},
+			},
+			expectedValid: true,
+		},
+		{
+			id: "valid secret with invalid path",
+			tests: []TestStepConfiguration{
+				{
+					As:                         "unit",
+					Commands:                   "commands",
+					ContainerTestConfiguration: &ContainerTestConfiguration{From: "ignored"},
+					Secret: Secret{
+						Name:      "secret",
+						MountPath: "path/to/secret",
+					},
+				},
+			},
+			expectedValid: false,
+		},
 	}
 
 	for _, tc := range testTestsCases {
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -282,6 +282,10 @@ type TestStepConfiguration struct {
 	// the repository root to _output/local/artifacts.
 	ArtifactDir string `json:"artifact_dir"`
 
+	// Secret is an optional secret object which
+	// will be mounted inside the test container.
+	Secret Secret `json:"secret,omitempty"`
+
 	// Only one of the following can be not-null.
 	ContainerTestConfiguration                      *ContainerTestConfiguration                      `json:"container,omitempty"`
 	OpenshiftAnsibleClusterTestConfiguration        *OpenshiftAnsibleClusterTestConfiguration        `json:"openshift_ansible,omitempty"`
@@ -291,6 +295,15 @@ type TestStepConfiguration struct {
 	OpenshiftAnsibleUpgradeClusterTestConfiguration *OpenshiftAnsibleUpgradeClusterTestConfiguration `json:"openshift_ansible_upgrade,omitempty"`
 	OpenshiftInstallerClusterTestConfiguration      *OpenshiftInstallerClusterTestConfiguration      `json:"openshift_installer,omitempty"`
 	OpenshiftInstallerSrcClusterTestConfiguration   *OpenshiftInstallerSrcClusterTestConfiguration   `json:"openshift_installer_src,omitempty"`
+}
+
+// Secret describes a secret to be mounted inside a test
+// container.
+type Secret struct {
+	// Secret name, used inside test containers
+	Name string `json:"name"`
+	// Secret mount path. Defaults to /usr/test-secret
+	MountPath string `json:"mount_path"`
 }
 
 // ContainerTestConfiguration describes a test that runs a
--- a/pkg/steps/pod.go
+++ b/pkg/steps/pod.go
@@ -17,12 +17,16 @@ import (
 	"github.com/openshift/ci-operator/pkg/junit"
 )
 
+const testSecretName = "test-secret"
+const testSecretDefaultPath = "/usr/test-secrets"
+
 type PodStepConfiguration struct {
 	As                 string
 	From               api.ImageStreamTagReference
 	Commands           string
 	ArtifactDir        string
 	ServiceAccountName string
+	Secret             api.Secret
 }
 
 type podStep struct {
@@ -54,35 +58,7 @@ func (s *podStep) Run(ctx context.Context, dry bool) error {
 	}
 	image := fmt.Sprintf("%s:%s", s.config.From.Name, s.config.From.Tag)
 
-	pod := &coreapi.Pod{
-		ObjectMeta: meta.ObjectMeta{
-			Name: s.config.As,
-			Labels: trimLabels(map[string]string{
-				PersistsLabel:    "false",
-				JobLabel:         s.jobSpec.Job,
-				BuildIdLabel:     s.jobSpec.BuildId,
-				ProwJobIdLabel:   s.jobSpec.ProwJobID,
-				CreatedByCILabel: "true",
-			}),
-			Annotations: map[string]string{
-				JobSpecAnnotation:                     s.jobSpec.RawSpec(),
-				annotationContainersForSubTestResults: s.name,
-			},
-		},
-		Spec: coreapi.PodSpec{
-			ServiceAccountName: s.config.ServiceAccountName,
-			RestartPolicy:      coreapi.RestartPolicyNever,
-			Containers: []coreapi.Container{
-				{
-					Name:                     s.name,
-					Image:                    image,
-					Command:                  []string{"/bin/sh", "-c", "#!/bin/sh\nset -eu\n" + s.config.Commands},
-					Resources:                containerResources,
-					TerminationMessagePolicy: coreapi.TerminationMessageFallbackToLogsOnError,
-				},
-			},
-		},
-	}
+	pod := s.generatePodForStep(image, containerResources)
 
 	// when the test container terminates and artifact directory has been set, grab everything under the directory
 	var notifier ContainerNotifier = NopNotifier
@@ -180,6 +156,7 @@ func TestStep(config api.TestStepConfiguration, resources api.ResourceConfigurat
 			From:        api.ImageStreamTagReference{Name: api.PipelineImageStream, Tag: string(config.ContainerTestConfiguration.From)},
 			Commands:    config.Commands,
 			ArtifactDir: config.ArtifactDir,
+			Secret:      config.Secret,
 		},
 		resources,
 		podClient,
@@ -196,5 +173,69 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
 		podClient:   podClient,
 		artifactDir: artifactDir,
 		jobSpec:     jobSpec,
+	}
+}
+
+func (s *podStep) generatePodForStep(image string, containerResources coreapi.ResourceRequirements) *coreapi.Pod {
+	pod := &coreapi.Pod{
+		ObjectMeta: meta.ObjectMeta{
+			Name: s.config.As,
+			Labels: trimLabels(map[string]string{
+				PersistsLabel:    "false",
+				JobLabel:         s.jobSpec.Job,
+				BuildIdLabel:     s.jobSpec.BuildId,
+				ProwJobIdLabel:   s.jobSpec.ProwJobID,
+				CreatedByCILabel: "true",
+			}),
+			Annotations: map[string]string{
+				JobSpecAnnotation:                     s.jobSpec.RawSpec(),
+				annotationContainersForSubTestResults: s.name,
+			},
+		},
+		Spec: coreapi.PodSpec{
+			ServiceAccountName: s.config.ServiceAccountName,
+			RestartPolicy:      coreapi.RestartPolicyNever,
+			Containers: []coreapi.Container{
+				{
+					Image:                    image,
+					Name:                     s.name,
+					Command:                  []string{"/bin/sh", "-c", "#!/bin/sh\nset -eu\n" + s.config.Commands},
+					Resources:                containerResources,
+					TerminationMessagePolicy: coreapi.TerminationMessageFallbackToLogsOnError,
+				},
+			},
+		},
+	}
+
+	if s.config.Secret.Name != "" {
+		pod.Spec.Containers[0].VolumeMounts = getSecretVolumeMountFromSecret(s.config.Secret.MountPath)
+		pod.Spec.Volumes = getVolumeFromSecret(s.config.Secret.Name)
+	}
+	return pod
+}
+
+func getVolumeFromSecret(secretName string) []coreapi.Volume {
+	return []coreapi.Volume{
+		{
+			Name: testSecretName,
+			VolumeSource: coreapi.VolumeSource{
+				Secret: &coreapi.SecretVolumeSource{
+					SecretName: secretName,
+				},
+			},
+		},
+	}
+}
+
+func getSecretVolumeMountFromSecret(secretMountPath string) []coreapi.VolumeMount {
+	if secretMountPath == "" {
+		secretMountPath = testSecretDefaultPath
+	}
+	return []coreapi.VolumeMount{
+		{
+			Name:      testSecretName,
+			ReadOnly:  true,
+			MountPath: secretMountPath,
+		},
 	}
 }
--- a/pkg/steps/pod_test.go
+++ b/pkg/steps/pod_test.go
@@ -182,3 +182,131 @@ func TestPodStepExecution(t *testing.T) {
 		})
 	}
 }
+
+func TestGetPodObjectMounts(t *testing.T) {
+	testCases := []struct {
+		name                 string
+		podStep              func(*podStep)
+		expectedVolumeConfig *v1.Pod
+	}{
+		{
+			name: "no secret name results in no mounted secrets",
+			podStep: func(expectedPodStepTemplate *podStep) {
+				expectedPodStepTemplate.config.Secret.Name = ""
+			},
+			expectedVolumeConfig: &v1.Pod{
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							VolumeMounts: []v1.VolumeMount{},
+						},
+					},
+					Volumes: []v1.Volume{},
+				},
+			},
+		},
+		{
+			name: "with secret name results in secret mounted with default path",
+			podStep: func(expectedPodStepTemplate *podStep) {
+				expectedPodStepTemplate.config.Secret.Name = testSecretName
+			},
+			expectedVolumeConfig: &v1.Pod{
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							VolumeMounts: []v1.VolumeMount{
+								{
+									Name:      testSecretName,
+									MountPath: testSecretDefaultPath,
+									ReadOnly:  true,
+								},
+							},
+						},
+					},
+					Volumes: []v1.Volume{
+						{
+							Name: testSecretName,
+							VolumeSource: v1.VolumeSource{
+								Secret: &v1.SecretVolumeSource{
+									SecretName: testSecretName,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "with secret name and path results in mounted secret with custom path",
+			podStep: func(expectedPodStepTemplate *podStep) {
+				expectedPodStepTemplate.config.Secret.Name = testSecretName
+				expectedPodStepTemplate.config.Secret.MountPath = "/usr/local/secrets"
+			},
+			expectedVolumeConfig: &v1.Pod{
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							VolumeMounts: []v1.VolumeMount{
+								{
+									Name:      testSecretName,
+									MountPath: "/usr/local/secrets",
+									ReadOnly:  true,
+								},
+							},
+						},
+					},
+					Volumes: []v1.Volume{
+						{
+							Name: testSecretName,
+							VolumeSource: v1.VolumeSource{
+								Secret: &v1.SecretVolumeSource{
+									SecretName: testSecretName,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			podStepTemplate := expectedPodStepTemplate()
+			tc.podStep(podStepTemplate)
+
+			pod := podStepTemplate.generatePodForStep("", v1.ResourceRequirements{})
+
+			if !equality.Semantic.DeepEqual(pod.Spec.Volumes, tc.expectedVolumeConfig.Spec.Volumes) {
+				t.Errorf("test %s failed. generated pod.Spec.Volumes was not as expected", tc.name)
+				t.Error(diff.ObjectReflectDiff(pod.Spec.Volumes, tc.expectedVolumeConfig.Spec.Volumes))
+			}
+			if !equality.Semantic.DeepEqual(pod.Spec.Containers[0].VolumeMounts, tc.expectedVolumeConfig.Spec.Containers[0].VolumeMounts) {
+				t.Errorf("test %s failed. generated pod.Spec.Container[0].VolumeMounts was not as expected", tc.name)
+				t.Error(diff.ObjectReflectDiff(pod.Spec.Containers[0].VolumeMounts, tc.expectedVolumeConfig.Spec.Containers[0].VolumeMounts))
+			}
+
+		})
+	}
+
+}
+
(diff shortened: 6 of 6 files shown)


pkg/defaults/defaults.go
```diff
@@ -144,7 +145,7 @@ func FromConfig(
 			releaseStep = release.AssembleReleaseStep(*rawStep.ReleaseImagesTagStepConfiguration, config.Resources, podClient, imageClient, artifactDir, jobSpec)
 
 		} else if rawStep.TestStepConfiguration != nil {
-			step = steps.TestStep(*rawStep.TestStepConfiguration, config.Resources, podClient, artifactDir, jobSpec)
+			step = steps.TestStep(*rawStep.TestStepConfiguration, config.Resources, podClient, artifactDir, jobSpec, testSecret)
```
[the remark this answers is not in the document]
droslean: It also needs to be validated. see [here](https://github.com/openshift/ci-operator/blob/master/pkg/api/config.go#L90-L112)

mjudeikis:
> but we have to inject the secret for each test from the ci-operator's configuration. 
See https://github.com/openshift/release/pull/2756/files#diff-dae6e8fc1689c5695304cb8a462cfb63R22
This is optional, but yes.

> Prowgen will have to be updated as well.
Please point to the location where updates are needed

> However, there are some security issues that we have to consider. By adding this feature, it will mean that any team can use anybody's secret and run whatever they want.

There is already ways to do this. Just spin a template and rsh to the pods and get what you need. There always be loopholes in the system. Long story short - we need this feature otherwise we will end up with 5+ more templates. And it would be easier just to move our jobs somewhere else :/

pkg/steps/pod.go
```diff
@@ -185,16 +151,91 @@ func TestStep(config api.TestStepConfiguration, resources api.ResourceConfigurat
...
+	return []coreapi.Volume{}
+}
+
+func podVolume(name string) []coreapi.VolumeMount {
+	if len(name) > 0 {
+		return []coreapi.VolumeMount{
+			{
+				Name:      "secret",
+				ReadOnly:  true,
+				MountPath: "/usr/secrets",
```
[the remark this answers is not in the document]
mjudeikis: I like delimiters idea better than a new flag. I don't expect this to be widely used either way.

pkg/defaults/defaults.go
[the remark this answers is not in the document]
mjudeikis: So you want this to be configurable instead of flag, or both?

droslean:
> Please point to the location where updates are needed

[Here](https://github.com/openshift/ci-operator-prowgen/blob/master/cmd/ci-operator-prowgen/main.go#L106) prowgen fills up the command arguments in the podspec. 

An example ci-operator config should look like:
```yaml
tests:
- as: unit
  commands: make unit
  secret: mysecret
  container:
    from: src
```
Then prowgen will get the `secret` field value and appends it in the arguments.

@petr-muller @stevekuznetsov @bbguimaraes  Any thoughts?

pkg/defaults/defaults.go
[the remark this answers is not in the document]
droslean: I would expect it to be defined in [here](https://github.com/openshift/ci-operator/blob/master/pkg/defaults/defaults.go#L409)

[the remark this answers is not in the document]
droslean: Of course, if we do this like this we won't have to make any updates in prowgen, since ci-operator will read the secret directly from its config. 

Let us decide then. @petr-muller @bbguimaraes @stevekuznetsov  Any thoughts on this one?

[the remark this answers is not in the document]
mjudeikis: Which make sense, as now you can define secret only for steps you need, and not for all of them. Limits attach surface.

petr-muller:
> @petr-muller @stevekuznetsov @bbguimaraes Any thoughts?

Haven't looked yet, looking now

pkg/steps/pod_test.go
```diff
@@ -182,3 +183,101 @@ func TestPodStepExecution(t *testing.T) {
...
+		},
+		{
+			name: "with test secret",
+			podStep: func(expectedPodStepTemplate *podStep) {
+				expectedPodStepTemplate.secret = "secret"
+			},
+			expectedPod: func(expectedPodTemplate *v1.Pod) {},
+		},
+	}
+	for _, tc := range testCases {
```
petr-muller: :+1: for the tests. Can we make them [subtests](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks)?

pkg/steps/pod.go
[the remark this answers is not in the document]
petr-muller: Personally, I don't mind about these hardcodings, they make usage simple, we do not need to expose every single element to be configurable.

I would like to see the `"secret"` name to be extracted to a constant (because it needs to match in `podSecret` and `podVolume` and it might not be clear to someone who reads this a month from now) and _maaaaybe_ call the mountpoint something less less generic (`ci-operator-secrets`? `configured-secrets`?) to reduce conflict possibility. But that's probably just bikeshedding.

mjudeikis:
After mongling with the code I would say - lets ditch the flag and do this:
```
// gathers artifacts from that step.
type TestStepConfiguration struct {
...
	Secret *Secret `json:"secret"`
...
}

// Secret describes test secret to be mounted inside test
// container.
type Secret struct {
	Name string `json:"name"`
	Path string `json:"path"`
}
```

Simple, testable, and scoped.

pkg/steps/pod_test.go
mjudeikis: are you using this pattern already somewhere?

stevekuznetsov: We should be everywhere, if not, congrats, lead the charge :)

mjudeikis:
@petr-muller @droslean check last commit I added. Removed flags, and left fully config based one. 
+split to separate secret object, as there might be more options in the future.
Tested locally too, works like a charm.

pkg/api/config.go
```diff
@@ -106,6 +106,18 @@ func validateTestStepConfiguration(fieldRoot string, input []TestStepConfigurati
 			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].commands: is required", fieldRoot, num))
 		}
 
+		if len(test.Secret.Name) > 0 {
+			if ok := regexp.MustCompile("^[a-zA-Z]*").MatchString(test.Secret.Name); !ok {
+				validationErrors = append(validationErrors, fmt.Errorf("%s[%d].name: '%s' secret name is not valid value, should be [a-zA-Z]", fieldRoot, num, test.Secret.Name))
+			}
+			// validate path only if name is passed
+			if len(test.Secret.Path) > 0 {
+				if ok := regexp.MustCompile("^/((\\/*)\\w+)+").MatchString(test.Secret.Path); !ok {
```
droslean: is `filepath.IsAbs(test.Secret.Path)` what we need here?

pkg/api/config_test.go
```diff
@@ -75,7 +75,7 @@ func TestValidateTests(t *testing.T) {
 			id: "test without `commands`",
 			tests: []TestStepConfiguration{
 				{
-					As: "test",
+					As:                         "test",
```
droslean: oops, you have go 1.11? The tests will fail. Please use `1.10`

pkg/api/types.go
```diff
@@ -293,6 +297,13 @@ type TestStepConfiguration struct {
 	OpenshiftInstallerSrcClusterTestConfiguration   *OpenshiftInstallerSrcClusterTestConfiguration   `json:"openshift_installer_src,omitempty"`
 }
 
+// Secret describes test secret to be mounted inside test
+// container.
+type Secret struct {
+	Name string `json:"name"`
+	Path string `json:"path"`
```
droslean: MountPath should be a better name.

pkg/steps/pod_test.go
```diff
@@ -182,3 +184,135 @@ func TestPodStepExecution(t *testing.T) {
...
+			podTemplate := expectedPodTemplate()
+			tc.podStep(podStepTemplate)
+			tc.expectedPod(podTemplate)
+			pod, err := podStepTemplate.getPodObject()
+			if err != nil {
+				t.Errorf("test case %s error %v", tc.name, err)
+			}
+
+			if !reflect.DeepEqual(pod, podTemplate) {
+				//if true {
```
droslean: no need for this comment.

pkg/steps/pod.go
```diff
@@ -198,3 +171,82 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
+
+func podVolume(config PodStepConfiguration) []coreapi.VolumeMount {
+	var secretPath string
+	if len(config.SecretName) > 0 {
+		if len(config.SecretPath) > 0 {
+			secretPath = config.SecretPath
+		} else {
+			secretPath = testSecretDefaultPath
+		}
+		return []coreapi.VolumeMount{
```
droslean: since we are mounting a single secret, I think we can use also the `volumeMounts.subPath`.

CONFIGURATION.md
```diff
@@ -313,6 +313,12 @@ of the images in the pipeline.
 ## `tests.container.from`
 `from` is the pipeline image tag that this test will be run on.
 
+## `tests.secret.name
```
stevekuznetsov: You're missing a grave accent at the end of this line

CONFIGURATION.md
```diff
@@ -313,6 +313,12 @@ of the images in the pipeline.
...
+`secret.name` is  secret name to be mounted inside a test container.
```
stevekuznetsov: ```is the name of the secret```?

CONFIGURATION.md
```diff
@@ -313,6 +313,12 @@ of the images in the pipeline.
...
+
+## `tests.secret.path
```
stevekuznetsov: Missing the trailing accent here too

CONFIGURATION.md
```diff
@@ -313,6 +313,12 @@ of the images in the pipeline.
...
+`secrets.path` is an optional variable for a secret mount path. Default value `/usr/secrets`
```
stevekuznetsov: `s/secrets.path/secret.path/`

stevekuznetsov: `"is the path at which to mount the secret. Optional, defaults to ..."`

pkg/api/config.go
```diff
@@ -106,6 +106,18 @@ func validateTestStepConfiguration(fieldRoot string, input []TestStepConfigurati
 			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].commands: is required", fieldRoot, num))
 		}
 
+		if len(test.Secret.Name) > 0 {
+			if ok := regexp.MustCompile("^[a-zA-Z]*").MatchString(test.Secret.Name); !ok {
+				validationErrors = append(validationErrors, fmt.Errorf("%s[%d].name: '%s' secret name is not valid value, should be [a-zA-Z]", fieldRoot, num, test.Secret.Name))
```
stevekuznetsov: Import the validator from upstream

pkg/api/config_test.go
stevekuznetsov: eek, we need to update

pkg/api/config_test.go
```diff
@@ -155,6 +155,64 @@ func TestValidateTests(t *testing.T) {
...
 		},
+		{
+			id: "invalid secret format",
+			tests: []TestStepConfiguration{
+				{
+					As:                                       "test",
+					OpenshiftAnsibleClusterTestConfiguration: &OpenshiftAnsibleClusterTestConfiguration{},
+					Secret: Secret{
+						Name: "secret",
+						Path: "/path/to/secret:exec",
```
stevekuznetsov: Is that not a valid filename on unix?

pkg/api/types.go
```diff
@@ -282,6 +282,10 @@ type TestStepConfiguration struct {
 	// the repository root to _output/local/artifacts.
 	ArtifactDir string `json:"artifact_dir"`
 
+	// Secret is an optional secret name which
+	// will be mounted inside the test container.
+	Secret Secret `json:"secret"`
```
stevekuznetsov: Should be omitempty

pkg/steps/pod.go
```diff
@@ -17,16 +17,22 @@ import (
...
 	Commands           string
 	ArtifactDir        string
 	ServiceAccountName string
+	SecretName         string
+	SecretPath         string
 }
 
 type podStep struct {
 	name        string
+	secret      string
```
stevekuznetsov: Why here and also under config?

pkg/steps/pod.go
```diff
@@ -44,44 +50,9 @@ func (s *podStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, er
 func (s *podStep) Run(ctx context.Context, dry bool) error {
 	log.Printf("Executing %s %s", s.name, s.config.As)
 
-	containerResources, err := resourcesFor(s.resources.RequirementsForStep(s.config.As))
```
stevekuznetsov: Moving all this makes it hard to see your diff, please don't do that

pkg/steps/pod.go
```diff
@@ -198,3 +171,82 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
+					TerminationMessagePolicy: coreapi.TerminationMessageFallbackToLogsOnError,
+					VolumeMounts:             podVolume(s.config),
+				},
+			},
+			Volumes: podSecret(s.config),
+		},
+	}, nil
+}
+
+func podSecret(config PodStepConfiguration) []coreapi.Volume {
```
stevekuznetsov: `s/podSecret/secretVolumeFromConfig/`?

pkg/steps/pod.go
```diff
@@ -198,3 +171,82 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
+						SecretName: config.SecretName,
+					},
+				},
+			},
+		}
+	}
+	return []coreapi.Volume{}
+}
+
+func podVolume(config PodStepConfiguration) []coreapi.VolumeMount {
```
stevekuznetsov: `s/podVolume
/secretVolumeMountFromConfig/`?

pkg/steps/pod.go
```diff
@@ -198,3 +171,82 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
+					VolumeMounts:             podVolume(s.config),
+				},
+			},
+			Volumes: podSecret(s.config),
+		},
+	}, nil
+}
+
+func podSecret(config PodStepConfiguration) []coreapi.Volume {
+	if len(config.SecretName) > 0 {
```
stevekuznetsov: Don't call unconditionally only to check if you should have been called at all. Above, do:

```go
pod := &coreapi.Pod{/* ... */}
if s.config.SecretName > 0 {
    pod.Spec.Containers[0].VolumeMounts = []coreapi.Volume{/* ... */}
}
```

pkg/steps/pod.go
```diff
@@ -198,3 +171,82 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
+			},
+		}
+	}
+	return []coreapi.Volume{}
+}
+
+func podVolume(config PodStepConfiguration) []coreapi.VolumeMount {
+	var secretPath string
+	if len(config.SecretName) > 0 {
+		if len(config.SecretPath) > 0 {
```
stevekuznetsov: More idiomatic in go to do:

```go
secretPath := config.SecretPath
if secretPath == "" {
    secretPath = testSecretDefaultPath
}
```

pkg/steps/pod_test.go
```diff
@@ -182,3 +184,135 @@ func TestPodStepExecution(t *testing.T) {
...
 }
+
+func TestGetPodObject(t *testing.T) {
+	testCases := []struct {
+		name        string
+		podStep     func(*podStep)
+		expectedPod func(*v1.Pod)
+	}{
+		{
+			name: "no secret",
```
stevekuznetsov: Give us cause and effect in test names please 

`"no secret name results in no mounted secrets"`

pkg/steps/pod_test.go
```diff
@@ -182,3 +184,135 @@ func TestPodStepExecution(t *testing.T) {
...
+			podStepTemplate := expectedPodStepTemplate()
+			podTemplate := expectedPodTemplate()
+			tc.podStep(podStepTemplate)
+			tc.expectedPod(podTemplate)
+			pod, err := podStepTemplate.getPodObject()
+			if err != nil {
+				t.Errorf("test case %s error %v", tc.name, err)
+			}
+
+			if !reflect.DeepEqual(pod, podTemplate) {
```
stevekuznetsov: Use `equality.Semantic.DeepEqual()` for k8s types

pkg/steps/pod_test.go
```diff
@@ -182,3 +184,135 @@ func TestPodStepExecution(t *testing.T) {
...
+				//if true {
+				t.Errorf("test %s failed. generated pod spec was not as expected", tc.name)
```
stevekuznetsov: Use `diff.ObjectReflectDiff` for diffing

pkg/steps/pod_test.go
```diff
@@ -182,3 +184,135 @@ func TestPodStepExecution(t *testing.T) {
 		})
 	}
 }
+
+func TestGetPodObject(t *testing.T) {
+	testCases := []struct {
+		name        string
+		podStep     func(*podStep)
+		expectedPod func(*v1.Pod)
```
stevekuznetsov: This might be easier to read as a `func(*v1.Pod, *testing.T)` and just have this func validate the pod itself

stevekuznetsov: Also technically you only validate the podspec, right?

pkg/steps/pod_test.go
```diff
@@ -182,3 +184,135 @@ func TestPodStepExecution(t *testing.T) {
...
+				t.Errorf("test %s failed. generated pod spec was not as expected", tc.name)
+				t.Error(pod)
+				t.Error(podTemplate)
+			}
+		})
+	}
+
+}
+
+func expectedPodTemplate() *v1.Pod {
```
stevekuznetsov: If you only check the fields you care about your test will not need this, will not be convoluted with other behavior of the system and will be more concise

pkg/api/config_test.go
mjudeikis: It is valid. but in some mount operations (example in docker run this is how you pass in option) it defines that mountpoint should be executable. Better safe than sorry.

pkg/api/config.go
mjudeikis: you need to be more specific here. What validator?

pkg/steps/pod_test.go
mjudeikis: We validate podSpec with PodTemplate which is mutated manually in the test scenario. So it validated function generated podSpec with our template. I think the test will become more complex if I stick with one function.

mjudeikis: I still need to "construct reference point" so second function is needed. I could remove all `func expectedPodTemplate() *v1.Pod` but this means this test will become unextendable and very narrow. With current implementation it can be extended to test any podSpec mutation.

pkg/steps/pod_test.go
mjudeikis: having a hard time to imagine how this could be squashed.

pkg/api/config_test.go
mjudeikis: This one is painful. Need to reconfigure my env to be dual-go-lang-version :////

pkg/steps/pod.go
```diff
@@ -198,3 +170,79 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
+	return pod, nil
+}
+
+func secretVolumeFromConfig(config PodStepConfiguration) []coreapi.Volume {
+	return []coreapi.Volume{
+		{
+			Name: testSecretName,
+			VolumeSource: coreapi.VolumeSource{
+				Secret: &coreapi.SecretVolumeSource{
+					SecretName: config.SecretName,
```
droslean: pass only the `secretName` instead of the whole `PodStepConfiguration`

pkg/steps/pod.go
```diff
@@ -198,3 +170,79 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
+				},
+			},
+		},
+	}
+}
+
+func secretVolumeMountFromConfig(config PodStepConfiguration) []coreapi.VolumeMount {
+	secretMountPath := config.SecretMountPath
+	if config.SecretMountPath == "" {
```
droslean: same here

pkg/api/config.go
droslean: `k8s.io/kubernetes/pkg/apis/core/validation`

pkg/steps/pod.go
mjudeikis: I prefer this to be more unified so both functions take the same input... But change as per suggestion.

pkg/api/config.go
mjudeikis: the price for this one was quite big. Added missing dependencies in the separate PR for easier review. Not sure if you want to do this for one regexp.

pkg/steps/pod_test.go
petr-muller: Looks like we do :)

```
$ git grep t.Run '*_test.go' ':!vendor'
pkg/api/config_test.go:         t.Run(testCase.name, func(t *testing.T) {
pkg/api/types_test.go:          t.Run(tt.name, func(t *testing.T) {
pkg/defaults/defaults_test.go:          t.Run(testCase.name, func(t *testing.T) {
pkg/steps/artifacts_test.go:            t.Run(tt.name, func(t *testing.T) {
pkg/steps/pod_test.go:          t.Run(tc.purpose, func(t *testing.T) {
```

stevekuznetsov:
What did you do to vendor? That diff doesn't look appropriate...

pkg/api/types.go
```diff
@@ -293,6 +297,13 @@ type TestStepConfiguration struct {
 	OpenshiftInstallerSrcClusterTestConfiguration   *OpenshiftInstallerSrcClusterTestConfiguration   `json:"openshift_installer_src,omitempty"`
 }
 
+// Secret describes a secret to be mounted inside a test
+// container.
+type Secret struct {
+	Name      string `json:"name"`
```
stevekuznetsov: This is where people will go after the README for docs on the fields, so let's make sure they are documented here

pkg/steps/pod.go
stevekuznetsov: Regardless -- break it out into a separate commit or don't do it, it's not germane to the changes you are making here.

mjudeikis:
> What did you do to vendor? That diff doesn't look appropriate...

add k8s validation, and glide install ... check the code for now, and I will try to drop it and redo...

pkg/steps/pod_test.go
stevekuznetsov: Instead of taking a template, mutating it and doing a DeepEqual, just validate the fields you need. Don't worry about this not being extensible, infinitely extensible unit tests aren't something we really need to be worrying about

pkg/steps/pod.go
```diff
@@ -198,3 +170,78 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
 		jobSpec:     jobSpec,
 	}
 }
+
+func (s *podStep) getPodObject() (*coreapi.Pod, error) {
```
stevekuznetsov: This could use a more descriptive name than `getPodObject`

pkg/steps/pod_test.go
stevekuznetsov: None of your tests care about the labels, or the annotations, or the command .. or ... or ... 

If you follow my suggestion and don't use `DeepEqual` but instead validate the specific fields you care about (feel free to `DeepEqual` on the  volume mounts, if you want) you don't need to present any extraneous fields and you don't need this func

pkg/steps/pod_test.go
```diff
@@ -182,3 +184,133 @@ func TestPodStepExecution(t *testing.T) {
...
+		name        string
+		podStep     func(*podStep)
+		expectedPod func(*v1.Pod)
+	}{
+		{
+			name: "no secret name results in no mounted secrets",
+			podStep: func(expectedPodStepTemplate *podStep) {
+				expectedPodStepTemplate.config.SecretName = ""
+			},
+			expectedPod: func(expectedPodTemplate *v1.Pod) {
```
stevekuznetsov: So for instance ... right now to understand this test case I need to understand the template and then understand how the mutation is working. Instead, you can do a in-place validation with a `func(*coreapi.Pod, *testing.T)`:

```go
expectedPod: func(input *coreapi.Pod, t *testing.T) {
    if len(input.Spec.Volumes) != 0 {
        t.Errorf("declared volumes when none were expected: %v", diff.ObjectReflectDiff([]coreapi.Volume{}, input.Spec.Volumes)
    }
    if len(input.Spec.Containers[0].VolumeMounts) != 0 {
        t.Errorf("mounted volumes when none were expected: %v", diff.ObjectReflectDiff([]coreapi.VolumeMount{}, input.Spec.Containers[0].VolumeMounts)
    }
}
```


expected

pkg/steps/pod_test.go
```diff
@@ -1,8 +1,10 @@
 package steps
 
 import (
+	"path/filepath"
 	"testing"
 
+	coreapi "k8s.io/api/core/v1"
```
stevekuznetsov: can you make sure you're always using `coreapi` and not `v1`?

pkg/steps/pod_test.go
mjudeikis: I like your minimalistic style, but at this place, we want to know if what we set is per expectation. Not only if it is set. Refactored to middle ground what was before and what are you suggesting.

pkg/api/config.go
droslean: hm. I would prefer the regex rather than vendoring the whole library. @petr-muller @stevekuznetsov any thoughts?

mjudeikis:
Glide is so breaking vendoring... :/ cant add one dependency without updating other... and you point to master `openshift/api` and `client-go` :/ still working

pkg/steps/pod.go
stevekuznetsov: `generatePodForStep`? `generatePodForConfig`?

pkg/steps/pod_test.go
```diff
@@ -182,3 +182,134 @@ func TestPodStepExecution(t *testing.T) {
...
+		name                 string
+		podStep              func(*podStep)
+		expectedVolumeConfig *v1.Pod
+	}{
+		{
+			name: "no secret name results in no mounted secrets",
+			podStep: func(expectedPodStepTemplate *podStep) {
+				expectedPodStepTemplate.config.SecretName = ""
+			},
+			expectedVolumeConfig: &v1.Pod{
```
stevekuznetsov: These look good

mjudeikis:
final changes are done, and squashed.

stevekuznetsov:
@petr-muller can you take a final pass here?

pkg/steps/pod.go
```diff
@@ -198,3 +182,65 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
+	}
+
+	if s.config.SecretName != "" {
+		pod.Spec.Containers[0].VolumeMounts = secretVolumeMountFromConfig(s.config.SecretMountPath)
+		pod.Spec.Volumes = secretVolumeFromConfig(s.config.SecretName)
+	}
+	return pod, nil
+}
+
+func secretVolumeFromConfig(secretName string) []coreapi.Volume {
```
droslean: nit: maybe the name `getVolumeFromSecret` or `getSecretVolume` ?

pkg/steps/pod.go
```diff
@@ -198,3 +182,65 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
+			VolumeSource: coreapi.VolumeSource{
+				Secret: &coreapi.SecretVolumeSource{
+					SecretName: secretName,
+				},
+			},
+		},
+	}
+}
+
+func secretVolumeMountFromConfig(secretMountPath string) []coreapi.VolumeMount {
```
droslean: nit: likewise here, maybe `getSecretVolumeMount`?

pkg/steps/pod.go
```diff
@@ -17,12 +17,17 @@ import (
...
+const testSecretName = "test-secret"
+const testSecretDefaultPath = "/usr/test-secrets"
+
 type PodStepConfiguration struct {
 	As                 string
 	From               api.ImageStreamTagReference
 	Commands           string
 	ArtifactDir        string
 	ServiceAccountName string
+	SecretName         string
```
droslean: why not pass the whole `Secret` struct here?

pkg/steps/pod.go
```diff
@@ -198,3 +182,65 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
+				{
+					Name:                     s.name,
+					Command:                  []string{"/bin/sh", "-c", "#!/bin/sh\nset -eu\n" + s.config.Commands},
+					TerminationMessagePolicy: coreapi.TerminationMessageFallbackToLogsOnError,
+				},
+			},
+		},
+	}
+
+	if s.config.SecretName != "" {
```
droslean: IMHO since we will always try to get both `volume` and `volumeMount`, I think it's better to have them in one function that returns both of values, and the nil check can be done by passing the whole `Secret` struct in the config.

pkg/steps/pod.go
stevekuznetsov: (lol that I suggested the other one)

pkg/steps/pod.go
stevekuznetsov: Will we always have both? I thought we convenience default the mount path

petr-muller:
I wasn't paying much attention to this, looked like @droslean is taking care of it. I feel like after ~90 comments it should be reviewed to death, don't block on me ;)

pkg/steps/pod.go
mjudeikis: I think it just overengineering. Normally we should always have both. But for this use-case. Future might change.

droslean: No problem from my side. That was just a nit.

stevekuznetsov:
Ah sorry
@droslean okay, your call

pkg/steps/pod.go
mjudeikis: Changed.

pkg/steps/pod.go
```diff
@@ -54,35 +58,12 @@ func (s *podStep) Run(ctx context.Context, dry bool) error {
...
-					Resources:                containerResources,
-					TerminationMessagePolicy: coreapi.TerminationMessageFallbackToLogsOnError,
-				},
-			},
-		},
+	pod, err := s.generatePodForStep()
+	if err != nil {
+		return err
 	}
+	pod.Spec.Containers[0].Image = image
```
droslean: nit: why not passing the `image` and `containerResources` in `generatePodForStep()`

droslean:
Please update also the https://github.com/openshift/ci-operator/blob/master/CONFIGURATION.md#ci-operator-configuration-reference

mjudeikis:
done

pkg/steps/pod.go
```diff
@@ -198,3 +178,67 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
 		jobSpec:     jobSpec,
 	}
 }
+
+func (s *podStep) generatePodForStep(image string, containerResources coreapi.ResourceRequirements) (*coreapi.Pod, error) {
```
droslean: here you return `error` but it will always be `nil` since you are not checking for any errors inside this function.

pkg/steps/pod.go
```diff
@@ -54,34 +58,9 @@ func (s *podStep) Run(ctx context.Context, dry bool) error {
...
-					Name:                     s.name,
-					Image:                    image,
-					Command:                  []string{"/bin/sh", "-c", "#!/bin/sh\nset -eu\n" + s.config.Commands},
-					Resources:                containerResources,
-					TerminationMessagePolicy: coreapi.TerminationMessageFallbackToLogsOnError,
-				},
-			},
-		},
+	pod, err := s.generatePodForStep(image, containerResources)
+	if err != nil {
```
droslean: this will always be nil

droslean:
/hold 
The pod will try to mount a secret that doesn't exist in the temporary namespace.  We have to update prowgen to add the `--secret-dir` arg and inject the secret in the prowjob's pod, or make the ci-operator to create the secret inside the temporary namespace.

I already discuss with @petr-muller , and we prefer the 2nd solution, as we don't want to involve `prowgen`

@stevekuznetsov @bbguimaraes @petr-muller  Any thoughts?

stevekuznetsov:
Why shouldn't we use the same secret copying mechanism as `--secret-dir`? It solves the same problem, right?

droslean:
`--secret-dir` requires the secrets to already be mounted in the pod (extra work for prowgen). But despite that, our case is simpler, since we want to mount a single secret. ci-operator can just get the secret from one namespace and create it to the temporary namespace.
example:
```yaml
secret:
  name: secret
  mount_path: /usr/secrets
  namespace: ci
```
Then ci-operator will get the secret from `ci` namespace and re-create it in the temporary namespace, then it will mount it in the specific test pod.

or
we merge that PR, and update the documentation that if anyone wants to use this feature he has to inject his secret in the prowjob and add the `--secret-dir` arg or make prowgen to do that.

I prefer the first sollution.  @stevekuznetsov thoughts?

stevekuznetsov:
Doing that would make it impossible to run `ci-operator` locally without access to read secrets in the `ci` namespace -- nobody has that. The `--secret-dir` impl is to make it so that the inputs for ci-operator can be provided literally.

/cc @smarterclayton

mjudeikis:
> Doing that would make it impossible to run `ci-operator` locally without access to read secrets in the `ci` namespace -- nobody has that. The `--secret-dir` impl is to make it so that the inputs for ci-operator can be provided literally.
> 
> /cc @smarterclayton

Agree. This is why I would say that adding this should be users responsibility (we already doing this). its one of those questions on how far we are willing to go :)

droslean:
Great. Let's also update the documentation in this PR.

smarterclayton:
In general the only secrets are the ones we manage, and profile is
supposed to be how we bundle them.  That keeps complexity low - single
input per build job.

We do need to think about how git secrets are provided, but that
wouldn’t be on the test, it would be in CI op and copied into the
namespace.

PR #256, opened 2019-02-07
Revert "test: Add an optional flag to the api for cluster tests"

smarterclayton:
This reverts commit 68dff40e6c01fd18d806ea4089e030f353ca7773.

I'm reverting this until we settle on what the actual supported set we want is.


smarterclayton:
@petr-muller @stevekuznetsov reverting until we decide on what we actually want to support is.

PR #257, opened 2019-02-08
When multiple containers report unit status, don't overlap time

smarterclayton:
Most of our test containers run either a single container (test)
or run multiple containers in succession (setup, test, teardown)
or have only a single container that is being reported (test, setup).

When testing pods in succession it's much more useful to show the
unique time per container (setup time, then test time, then teardown
time) than to show their full duration. In the future we could add
an annotation to tune this behavior, but for now sorting and then
showing non-overlapping time is more useful.

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -10,8 +10,10 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sort"
 	"strings"
 	"sync"
+	"time"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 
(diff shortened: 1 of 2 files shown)


smarterclayton:
Testing this in the real cluster (trying to quantify how slow tests and teardown are.

smarterclayton:
And yes, I self tagged, and yes I know that's bad.  I may revert this if it doesn't give me the results I need.

pkg/steps/artifacts_test.go
```diff
@@ -252,13 +250,188 @@ func TestTestCaseNotifier_SubTests(t *testing.T) {
...
+									FinishedAt: meta.Time{Time: time.Unix(1150, 0)},
+								},
+							},
+						},
+					},
+				},
+			},
+			wantTests: []*junit.TestCase{
+				{Name: "container other", Duration: 100},
+				{Name: "container test", FailureOutput: &junit.FailureOutput{Output: "exit message"}, Duration: 0},
```
stevekuznetsov: Why is this useful for a user

pkg/steps/artifacts_test.go
```diff
@@ -252,13 +250,188 @@ func TestTestCaseNotifier_SubTests(t *testing.T) {
...
+									FinishedAt: meta.Time{Time: time.Unix(1100, 0)},
+								},
+							},
+						},
+					},
+				},
+			},
+			wantTests: []*junit.TestCase{
+				{Name: "container other", Duration: 100},
+				{Name: "container test", FailureOutput: &junit.FailureOutput{Output: "exit message"}, Duration: 0},
```
stevekuznetsov: Why is this desired

pkg/steps/artifacts.go
```diff
@@ -107,24 +109,47 @@ func (n *TestCaseNotifier) SubTests(prefix string) []*junit.TestCase {
 	if len(names) == 0 {
 		return nil
 	}
+	statuses := make([]coreapi.ContainerStatus, len(pod.Status.ContainerStatuses))
```
stevekuznetsov: I think it might be nice for code health to put the digestion of ContainerStatus --> timelines into a separate func and have some unit tests for it

pkg/steps/artifacts_test.go
smarterclayton: Having multiple overlapping containers running at the same time means you want "linear" flow.  If we suddenly flip to "parallel" flow just because of a short test, that means you can't predict what you would get.  So we are effectively treating > 1 containers as always linear flow.

smarterclayton: Basically:

```
When testing pods in succession it's much more useful to show the
unique time per container (setup time, then test time, then teardown
time) than to show their full duration. In the future we could add
an annotation to tune this behavior, but for now sorting and then
showing non-overlapping time is more useful.
```

We only have three cases - single container, two containers where only one is identified in the annotation, and multiple containers in the annotation.  The behavior single and two want is the same as linear time with a single container.

PR #261, opened 2019-02-12
Add installer container test

pmorie:
Add support for a test like `openshift_installer`, but that allows you to specify a pipeline image with `from`

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -289,14 +289,15 @@ type TestStepConfiguration struct {
 	ArtifactDir string `json:"artifact_dir"`
 
 	// Only one of the following can be not-null.
-	ContainerTestConfiguration                      *ContainerTestConfiguration                      `json:"container,omitempty"`
-	OpenshiftAnsibleClusterTestConfiguration        *OpenshiftAnsibleClusterTestConfiguration        `json:"openshift_ansible,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/api/types.go
```diff
@@ -385,6 +386,16 @@ type OpenshiftInstallerSrcClusterTestConfiguration struct {
 	ClusterTestConfiguration `json:",inline"`
 }
 
+// OpenshiftInstallerContainerClusterTestConfiguration describes a test that
+// provisions a cluster using openshift-installed and executes a command in a
```
jboyd01: typo   openshift-installer

pmorie: Thanks

smarterclayton:
What is this trying to do?  Please add an example of what your test is going to be.

pmorie:
@smarterclayton The goal is to enable e2e to run using a container image built in the ci-operator pipeline. For example, say that I have a health-check image built in my pipeline and I want to run commands in that image against software deployed by my operator:

```yaml
- as: olm-e2e
  # The 'olm-e2e' test installs federation-v2 into a test cluster using OLM, and
  # drives an E2E test using the installed federation.
  commands: |
    ./healthcheck-my-operator.sh
  openshift_installer_container:
    cluster_profile: aws
    from: operator-health-check
```

PR #263, opened 2019-02-15
api: Support a `memory_backed_volume` for tests that need fast IO

smarterclayton:
Some tests like the OpenShift integration tests are extremely IO
intensive in a way that can disrupt other jobs. Allow these jobs to
request a memory backed filesystem to be mounted into the test
container at a predefined location.

```
tests:
- as: foo
  command: bar
  container:
    memory_backed_volume:
      size: 1Gi
```

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -312,12 +312,25 @@ type Secret struct {
 	MountPath string `json:"mount_path"`
 }
 
+// MemoryBackedVolume describes a tmpfs (memory backed volume)
+// that will be mounted into a test container at /tmp/volume.
+// Use with tests that need extremely fast disk, such as those
+// that run an etcd server or other IO-intensive workload.
+type MemoryBackedVolume struct {
+	// Size is the requested size of the volume as a Kubernetes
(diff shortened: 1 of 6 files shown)


pkg/api/types.go
```diff
@@ -312,12 +312,25 @@ type Secret struct {
...
+
 // ContainerTestConfiguration describes a test that runs a
 // command in one of the previously built images.
 type ContainerTestConfiguration struct {
 	// From is the image stream tag in the pipeline to run this
 	// command in.
 	From PipelineImageStreamTagReference `json:"from"`
+	// MemoryBackedVolume mounts a volume of the specified size into
+	// the container at /tmp/volume.
+	MemoryBackedVolume *MemoryBackedVolume `json:"memory_backed_volume"`
```
stevekuznetsov: Omitempty?

smarterclayton: do we have any others?  I thought we decided not to use them.

stevekuznetsov: Did we? I thought we needed it to keep parsing working. Oh but you made it a pointer. Either works

PR #269, opened 2019-02-17
Correctly handle empty input on source when nothing has to be changed

smarterclayton:
xargs fails if input is empty, but we want it to do nothing.


smarterclayton:
@stevekuznetsov some builds are failing

smarterclayton:
hive fails as do a few others

smarterclayton:
Merging to unblock the UBI switch

PR #270, opened 2019-02-17
We need to be able to run tests against many releases

smarterclayton:
'origin' is not a magic constant, remove the check for a known type
of release.

Blocks moving content to ocp namespace in preparation for 4.0 release
(openshift/origin-v4.0 is going to be the external location)

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -11,8 +11,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
-var originReleaseTagRegexp = regexp.MustCompile(`^origin-v\d+\.\d+$`)
-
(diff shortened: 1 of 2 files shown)


smarterclayton:
Prereq for openshift/release#287 which I am trying to land tomorrow.

New rules are 'ocp' is where all CI for origin will be published to, then there will be other scripts that clone it back.  We can add more restricted validation later.

bbguimaraes:
The reason for that check was to restrict the format of this string when generating jobs that require release RPMs:

https://github.com/openshift/ci-operator-prowgen/blob/bab4fd445b092a9bfa86e398d39869e5bfbf665d/cmd/ci-operator-prowgen/main.go#L218

I don't know how valuable it actually was.

PR #277, opened 2019-02-21
Update serialization code

stevekuznetsov:
Use YAML annotations for config serialization

Turns out you can't YAML dump without these.

---

Make optional fields omitempty

This way we don't dump them when they're empty

---

/assign @smarterclayton 
/cc @petr-muller @droslean @bbguimaraes

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/openshift/ci-operator/pkg/defaults"
+	"github.com/openshift/ci-operator/pkg/load"
 
 	coreapi "k8s.io/api/core/v1"
 	rbacapi "k8s.io/api/rbac/v1"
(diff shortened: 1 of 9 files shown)


smarterclayton:
Any chance something is using json still?  Should work the same, but might impact deserialization

smarterclayton:
Looks like this changes nil serialization?

stevekuznetsov:
Yeah I wanted to `omitempty` some stuff so the generated configs wouldn't have nil fields everywhere.

This seems to break `yaml.Unmarshal` somehow...
```
ERRO[0000] Failed to load CI Operator configuration      error="invalid ci-operator config: configuration has 3 errors:\n\n  * 'build_root' is required and must have image_stream_tag or project_image\n  * tests[0] has no type, you may want to specify 'container' for a container based test\n  * tests[1] has no type, you may want to specify 'container' for a container based test\n" source-file=/home/skuznets/releasego/src/github.com/openshift/release/ci-operator/config/containers/libpod/containers-libpod-master.yaml
```

Nothing is using JSON loads anymore

petr-muller:
I hate how this passes all tests :/

stevekuznetsov:
So it works ... then I change to YAML anchors and it breaks. What the fuck?

droslean:
The problem is in the library `github.com/ghodss/yaml`. If you try to use directly the `gopkg.in/yaml.v2` everything works as normal. 

There must be a bug in `github.com/ghodss/yaml`. I will try to investigate.

droslean:
oh lol.. I now realize that the tests are failing because they are in `json`

petr-muller:
> We were using `github.com/ghodss/yaml` to support json to yaml, right? Now we have everything in `yaml`, why not just using `gopkg.in/yaml.v2` directly instead? (note: `github.com/ghodss/yaml` is using `yaml.v2` library as well)

I think I was the one who introduced using `ghodss/yaml` lib back when we started supporting YAML. The reason to use it is not to support JSON (any YAML parser should be able to parse JSON, because it's a subset of YAML). Reason 1 for using it: It is also used by Prow, which I used to piggyback on. Reason 2 for using it: it **reuses** the `json` struct tags (it does not use `yaml` ones) - that was a good thing back then, because that way I did not need to change the struct tags, I only needed to change the methods ci-operator uses for (un)marshaling.

One reason why stuff broke could be the commit that changed the struct tags to `yaml`. Another could be some trickiness with using anchors that breaks the way the `ghodss/yaml` works:

_As a result, a completely acceptable way to handle YAML here is to convert it to JSON first and then use json.Unmarshal to turn it into a struct. And to serialize into YAML, to use json.Marshal to convert a struct to JSON and then convert that JSON to YAML. Fortunately, the go-yaml library works just fine for this task, and only needs some code to use it to convert YAML to JSON instead of serialize/deserialize structs._ ([source](http://ghodss.com/2014/the-right-way-to-handle-yaml-in-golang/))

It seems that if we 1) have no intention to support JSON marshal for any reason and 2) fix the struct tags, using the `gopkg.in/yaml.v2` might be the way to go if @droslean says using it directly solves the issue.

stevekuznetsov:
OK, cut out the other changes. This is ready/

PR #278, opened 2019-02-21
Require full branch name with --git-ref

jlebon:
This bit me in https://github.com/openshift/origin/pull/22081. My PR
branches are all of the form `pr/$name`, but using
`--git-ref jlebon/origin@$name` still worked because `git ls-remote`
just needs a trailing match. However, it will then throw off the image
builds in OCP.

As a sanity check, if the ref is a regular branch and not e.g. a tag, or
pull request, let's verify that the full branch name was specified.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -857,9 +857,18 @@ func jobSpecFromGitRef(ref string) (*api.JobSpec, error) {
 	if err != nil {
 		return nil, fmt.Errorf("'git ls-remote %s %s' failed with '%s'", repo, parts[1], err)
 	}
-	sha := strings.Split(strings.Split(string(out), "\n")[0], "\t")[0]
+	resolved := strings.Split(strings.Split(string(out), "\n")[0], "\t")
+	sha := resolved[0]
 	if len(sha) == 0 {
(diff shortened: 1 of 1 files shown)


droslean:
I am not sure what we want to accomplish here. There is already a check [here](https://github.com/openshift/ci-operator/blob/master/cmd/ci-operator/main.go#L847-L854) that it will never reach the code added in this PR.

jlebon:
> I am not sure what we want to accomplish here. There is already a check [here](https://github.com/openshift/ci-operator/blob/master/cmd/ci-operator/main.go#L847-L854) that it will never reach the code added in this PR.

This is orthogonal. Those checks you pointed out simply verify that the argument follows `ORG/NAME@REF`. The check we're adding here verifies that `REF` actually points to a full branch name, rather than just partially. `git ls-remote` just does a "suffix" match rather than a full ref match (try e.g. `git ls-remote https://github.com/openshift/origin.git head | less`). So we may be allowing refs that actually are not complete, and will cause failures down the line (see the linked issue for an example).

stevekuznetsov:
@jlebon sorry, just saw this

/approve
/lgtm

I like the check

PR #280, opened 2019-02-27
source: Periodic jobs may have nil refs

smarterclayton:
We should build source from the first available repo or default to
/go if necessary.

https://openshift-gce-devel.appspot.com/build/origin-ci-test/logs/periodic-ci-origin-conformance-k8s/231/


smarterclayton:
Fixed the linting error, ptal

smarterclayton:
@petr-muller @bbguimaraes if you have a chance to look

PR #281, opened 2019-02-28
Make canonical go repository omitempty

petr-muller:
I'm not sure why this was not present in https://github.com/openshift/ci-operator/pull/277, but the field really looks optional (and many existing configs do not set it). Noticed this while working on https://github.com/openshift/ci-operator-prowgen/pull/98/

/cc @stevekuznetsov


stevekuznetsov:
/lgtm

Thanks, overlooked this one

PR #283, opened 2019-03-03
Remove support for release configurations not in a single stream

smarterclayton:
We don't use this code and are unlikely to at any point in the future.
Clean up these code paths prior to supporting upgrade logic.

Builds on #282, only the latest commit is new

### DIFF
--- a/pkg/steps/release/promote.go
+++ b/pkg/steps/release/promote.go
@@ -31,10 +31,7 @@ type promotionStep struct {
 }
 
 func targetName(config api.PromotionConfiguration) string {
-	if len(config.Name) > 0 {
-		return fmt.Sprintf("%s/%s:${component}", config.Namespace, config.Name)
-	}
-	return fmt.Sprintf("%s/${component}:%s", config.Namespace, config.Tag)
+	return fmt.Sprintf("%s/%s:%s", config.Namespace, config.Name, api.ComponentFormatReplacement)
 }
 
 func (s *promotionStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, error) {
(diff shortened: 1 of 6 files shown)


pkg/steps/release/promote.go
```diff
@@ -71,107 +68,49 @@ func (s *promotionStep) Run(ctx context.Context, dry bool) error {
...
-			if _, err := s.dstClient.ImageStreams(s.config.Namespace).Update(is); err != nil {
-				if errors.IsConflict(err) {
-					return err
-				}
-				return fmt.Errorf("could not promote image streams: %v", err)
-			}
-			return nil
-		})
+	if len(s.config.Name) == 0 {
+		return fmt.Errorf("name is a required field for release tagging")
```
stevekuznetsov: We should do this on config load

pkg/steps/release/release_images.go
```diff
@@ -122,10 +122,7 @@ func (s *releaseImagesTagStep) Inputs(ctx context.Context, dry bool) (api.InputD
 }
 
 func sourceName(config api.ReleaseTagConfiguration) string {
-	if len(config.Name) > 0 {
-		return fmt.Sprintf("%s/%s:${component}", config.Namespace, config.Name)
-	}
-	return fmt.Sprintf("%s/${component}:%s", config.Namespace, config.Tag)
```
stevekuznetsov: Didn't you just make a type for `${component}`

pkg/steps/release/promote.go
smarterclayton: I added it in the validation as well, I just wanted to make sure we had something here in case someone came in and got confused.

pkg/steps/release/release_images.go
smarterclayton: global constant yes.

smarterclayton:
Remove `tag` you mean?

smarterclayton:
I verified no one is using the tag form, removed from API, updated docs, code, and tests.

pkg/steps/release/release_images.go
smarterclayton: Fixed

smarterclayton:
PTAL

PR #284, opened 2019-03-04
Create an initial release payload as `release:initial`

smarterclayton:
When we tag images in, create a second image stream `stable-initial` that
will not be modified and then let a secondary task create an image stream
from that alternate source. If RELEASE_IMAGE_INITIAL is already specified,
skip the action and instead tag it to that location.

Makes it easy to test an upgrade from the previous state to the current
state (post change) while also allowing for injecting an initial step.

Built on top of #282 #283

### DIFF
--- a/pkg/steps/release/create_release.go
+++ b/pkg/steps/release/create_release.go
@@ -4,6 +4,12 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
+	"strings"
+	"time"
+
+	coreapi "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	imageapi "github.com/openshift/api/image/v1"
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
@@ -20,6 +26,7 @@ import (
 // created, then invoking the admin command for building a new release.
 type assembleReleaseStep struct {
 	config      api.ReleaseTagConfiguration
+	latest      bool
 	resources   api.ResourceConfiguration
 	imageClient imageclientset.ImageV1Interface
 	podClient   steps.PodClient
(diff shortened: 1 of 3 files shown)


smarterclayton:
@stevekuznetsov @bbguimaraes @petr-muller this is a prerequisite for the upgrade test framework.  See openshift/release#2997 for an example of what the prow job looks like.  We probably need to add support for a new flag, and we need to keep in mind that the PR testing is going to have future variants (i.e. now it's release tag -> tag + PR images, but in future will include "last stable version" and "upgrade then downgrade" testing), so we don't want to assume it'll be a simple variant yet.

Every operator repo will get one of these, but it will be optional / not triggered by default.  That means we need to solve how to generate optional prow job status so we can stop manually crafting prow jobs for those.

smarterclayton:
Rebased on top of the merged PRs.  One more follow up coming (that will build on this) to default the contents of `stable` from `RELEASE_IMAGE_LATEST` when it is specified.

pkg/steps/release/create_release.go
```diff
@@ -103,16 +151,35 @@ func (s *assembleReleaseStep) Done() (bool, error) {
 }
 
 func (s *assembleReleaseStep) Requires() []api.StepLink {
-	return []api.StepLink{api.ImagesReadyLink()}
+	// if our prereq is provided, we don't need any prereqs
+	if len(os.Getenv(s.envVar())) > 0 {
```
stevekuznetsov: We should use `os.LookupEnv` for this?

pkg/steps/release/create_release.go
```diff
@@ -125,22 +192,31 @@ func (s *assembleReleaseStep) Provides() (api.ParameterMap, api.StepLink) {
...
 			}
-			return fmt.Sprintf("%s:%s", registry, "latest"), nil
+			return fmt.Sprintf("%s:%s", registry, tag), nil
 		},
-	}, api.ReleasePayloadImageLink(api.PipelineImageStreamTagReference("latest"))
+	}, api.ReleasePayloadImageLink(api.PipelineImageStreamTagReference(tag))
 }
 
-func (s *assembleReleaseStep) Name() string { return "[release:latest]" }
+func (s *assembleReleaseStep) Name() string {
```
stevekuznetsov:
```go
func (s *assembleReleaseStep) Name() string {
    return fmt.Sprintf("[release:%s]", s.tag())
}
```

pkg/steps/release/create_release.go
```diff
@@ -125,22 +192,31 @@ func (s *assembleReleaseStep) Provides() (api.ParameterMap, api.StepLink) {
...
+	}
+	return "[release:initial]"
+}
 
 func (s *assembleReleaseStep) Description() string {
-	return fmt.Sprintf("Create a release image in the release image stream")
+	if s.latest {
+		return "Create the release image containing all images built by this job"
+	}
+	return "Create the release image for the initial state of this job"
```
stevekuznetsov: initial state of current release at the time the job ran? could be more clear

pkg/steps/release/create_release.go
smarterclayton: yeah, I'm still in the older habit.

pkg/steps/release/create_release.go
smarterclayton: ok

pkg/steps/release/create_release.go
smarterclayton: Actually, the value has to be greater than zero.  we ignore `RELEASE_IMAGE_INITIAL=` because having an empty env is invalid and if someone did bash to init the env var we want that.  We don't discriminate between empty and unset here.

smarterclayton:
Updated PTAL

PR #286, opened 2019-03-04
Remove support for release configurations not in a single stream

smarterclayton:
We don't use this code and are unlikely to at any point in the future.
Clean up these code paths prior to supporting upgrade logic.

Preserve promoting to image repositories for things like ci-operator.


smarterclayton:
@stevekuznetsov sorry, I missed this one.

stevekuznetsov:
Ah, makes sense
/lgtm

PR #287, opened 2019-03-05
Default `stable` image stream from RELEASE_IMAGE_LATEST

smarterclayton:
Periodic jobs that want to run a PR from a known `RELEASE_IMAGE_LATEST`
need access to the `cli` and `installer` images out of the payload. This
fixes problems we have testing older installers and cli binaries against
newer payloads and fixes the current break in the release controllers. This
is also required for properly testing the installer.

Use standard oc commands (that are API forward compatible) to extract the
`cli` image and the `installer` image from the payload and set them on
stable if RELEASE_IMAGE_LATEST is provided as an input (as they are for the
release blocking jobs) or on stable-initial if RELEASE_IMAGE_INITIAL is
provided.

Hide nested container output because it is confusing when leveraging a pod
from within an existing step.

Builds on #284

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -15,13 +15,14 @@ import (
 	"sync"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/sets"
+	"github.com/golang/glog"
 
 	coreapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes/scheme"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
@@ -182,7 +183,7 @@ type PodClient interface {
 }
 
 func copyArtifacts(podClient PodClient, into, ns, name, containerName string, paths []string) error {
-	log.Printf("Copying artifacts from %s into %s", name, into)
+	glog.V(4).Infof("Copying artifacts from %s into %s", name, into)
 	var args []string
 	for _, s := range paths {
 		args = append(args, "-C", s, ".")
@@ -253,7 +254,10 @@ func copyArtifacts(podClient PodClient, into, ns, name, containerName string, pa
 		size += h.Size
 	}
 
-	if size > 0 {
+	// If we're updating a substantial amount of artifacts, let the user know as a way to
+	// indicate why the step took a long amount of time. Conversely, if we just got a small
+	// number of files this is just noise and can be omitted to not distract from other steps.
+	if size > 1*1000*1000 {
 		log.Printf("Copied %0.2fMi of artifacts from %s to %s", float64(size)/1000000, name, into)
 	}
 
@@ -650,6 +654,9 @@ func gatherContainerLogsOutput(podClient PodClient, artifactDir, namespace, podN
 	list, err := podClient.Pods(namespace).List(meta.ListOptions{FieldSelector: fields.Set{"metadata.name": podName}.AsSelector().String()})
 	if err != nil {
 		return fmt.Errorf("could not list pod: %v", err)
+	}
+	if len(list.Items) == 0 {
+		return nil
 	}
 	pod := &list.Items[0]
 
--- a/pkg/steps/pod.go
+++ b/pkg/steps/pod.go
@@ -21,7 +21,15 @@ import (
 const testSecretName = "test-secret"
 const testSecretDefaultPath = "/usr/test-secrets"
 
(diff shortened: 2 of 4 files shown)


pkg/steps/artifacts.go
```diff
@@ -253,7 +254,7 @@ func copyArtifacts(podClient PodClient, into, ns, name, containerName string, pa
 		size += h.Size
 	}
 
-	if size > 0 {
+	if size > 1*1000*1000 {
```
stevekuznetsov: Why? Document tpls

pkg/steps/artifacts.go
```diff
@@ -651,6 +652,9 @@ func gatherContainerLogsOutput(podClient PodClient, artifactDir, namespace, podN
 	if err != nil {
 		return fmt.Errorf("could not list pod: %v", err)
 	}
+	if len(list.Items) == 0 {
```
stevekuznetsov: Why are we doing a list with a field selector instead of a get?

pkg/steps/release/create_release.go
```diff
@@ -2,12 +2,17 @@ package release
...
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/sets"
```
stevekuznetsov: nit: in the larger stanza pls

pkg/steps/pod.go
```diff
@@ -48,8 +49,9 @@ func (s *podStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, er
 }
 
 func (s *podStep) Run(ctx context.Context, dry bool) error {
-	log.Printf("Executing %s %s", s.name, s.config.As)
-
+	if !s.config.SkipLogs {
```
stevekuznetsov: This smells bad, why isn't this going to make for inscrutable logs that admins hate?

pkg/steps/release/create_release.go
```diff
@@ -145,6 +121,207 @@ oc adm release extract --from=%q --to=/tmp/artifacts/release-payload-%s
 	return step.Run(ctx, dry)
 }
 
+// importFromReleaseImage uses the provided release image and updates the stable / release streams as
+// appropriate with the contents of the payload so that downstream components are using the correct images.
+// The most common case is to use the correct installer image, tests, and cli commands.
+func (s *assembleReleaseStep) importFromReleaseImage(ctx context.Context, dry bool, providedImage string) error {
```
stevekuznetsov: I don't think I fully understand why this is necessary, is this lacking in current version of ci-op or is it just when you want to run a job and specify a specific release?

pkg/steps/artifacts.go
smarterclayton: Reduce pointless output.  Will comment.  The only reason to print it is when we upload lots of content and the job just sits there.

pkg/steps/pod.go
smarterclayton: There are pod steps called within jobs (where you want artifacts).  The log output when you call these nested logs is more confusing - the step for "build release image" doesn't need to show `executing pod release-latest` because it already said what it is doing.  We still print pod logs when they fail which is 95% of all debugging.

pkg/steps/release/create_release.go
smarterclayton: We want to run a ci-operator job where the input to the job is a release image (release qualification tests) or we want to run an upgrade job where we take a previously released image (upgrade tests) where we won't have an image stream to represent them.  The upgrade job:

> Test the upgrade from 4.0.3 to a PR in 4.1 master against the cluster-version-operator component

wants to use the installer image from 4.0.3, so we need to populate `stable-initial` with the contents of 4.0.3 while `stable` continues to be populated from 4.1 master.

pkg/steps/artifacts.go
smarterclayton: That is an excellent question.   Will fix.

pkg/steps/pod.go
smarterclayton: The import from initial needs to run two pods, but the pods should never fail and the output of the nested pod within the import step is very confusing when reading it.  Only the top most step matters in terms of you needing to understand what is happening.

pkg/steps/release/create_release.go
smarterclayton: The other use case is:

> Test installing 4.1.4 and run jobs against it

We don't want to use the image stream for `4.1.4`, we want to grab the payload for `4.1.4` and then use the `installer` image out of the `4.1.4` payload to install the cluster and the `tests` image out of the `4.1.4` payload (in this case) to run the e2e tests.

pkg/steps/pod.go
smarterclayton: Added a long godoc on PodStepConfiguration to explain this, let me know if that helps clarify.

pkg/steps/release/create_release.go
smarterclayton: I added a super long godoc to the type, hope that clarifies.

smarterclayton:
Added requested comments, tried to clarify intent better.

PR #288, opened 2019-03-05
Implement a Golang cluster installer job that hides the template

smarterclayton:
Take the existing cluster-launch-installer-e2e template and hide it inside
a e2e step definition. Support an `upgrade` flag that alters the parameters
passed to the template such that the installer image from the input (either
release tag or an actual payload) is used to install, then launch the
upgrade tests from the job definition.

This sets the path for moving this functionality into the operator.

The user still has to upload the right cluster profile secret, but now we
print an error if we don't find it (for stronger typing). Regular users can
launch an upgrade from a single command.

Builds off #287


ARCHITECTURE.md
```diff
@@ -133,12 +133,7 @@ output, with parallel execution of steps leading to interwoven log output:
 2018/10/08 05:29:15 Building machine-config-operator
 2018/10/08 05:29:15 Building machine-config-controller
 2018/10/08 05:31:47 Build machine-config-server succeeded after 2m28s
-2018/10/08 05:31:47 Tagging machine-config-server into stable
```
stevekuznetsov: Why?

cmd/ci-operator/main.go
```diff
@@ -665,9 +665,6 @@ func (o *options) initializeNamespace() error {
 		})
 	}
 
-	if len(o.secrets) > 0 {
-		log.Printf("Populating secrets for test")
```
stevekuznetsov: Why?

pkg/steps/clusterinstall/clusterinstall.go
```diff
@@ -0,0 +1,140 @@
...
+				}
+			}
+		}
+		params = api.NewOverrideParameters(params, overrides)
+	}
+
+	step := steps.TemplateExecutionStep(template, params, podClient, templateClient, artifactDir, jobSpec)
+	subTests, ok := step.(nestedSubTests)
+	if !ok {
+		panic(fmt.Sprintf("unexpected %T", step))
```
stevekuznetsov: this sounds fun

pkg/steps/clusterinstall/clusterinstall.go
```diff
@@ -0,0 +1,140 @@
...
+		nestedSubTests: subTests,
+	}
+}
+
+func (s *e2eTestStep) checkPrereqs() error {
+	return nil
+}
+
+func (s *e2eTestStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, error) {
+	return nil, nil
```
stevekuznetsov: ???

pkg/steps/clusterinstall/clusterinstall.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (s *e2eTestStep) Run(ctx context.Context, dry bool) error {
+	if _, err := s.secretClient.Secrets(s.jobSpec.Namespace).Get(fmt.Sprintf("%s-cluster-profile", s.testConfig.As), meta.GetOptions{}); err != nil {
+		return fmt.Errorf("could not find required secret: %v", err)
+	}
+	return s.step.Run(ctx, dry)
+}
+
+func (s *e2eTestStep) Done() (bool, error) {
+	return false, nil
```
stevekuznetsov: ???

pkg/steps/template.go
```diff
@@ -81,7 +82,9 @@ func (s *templateExecutionStep) Run(ctx context.Context, dry bool) error {
 		}
 	}
 
-	addArtifactsToTemplate(s.template)
+	if len(s.artifactDir) > 0 {
```
stevekuznetsov: Should this have failed a test somewhere?

stevekuznetsov:
This doesn't seem finished

cmd/ci-operator/main.go
smarterclayton: Not actually useful, was just noise.  We already print when we populate a secret.

ARCHITECTURE.md
smarterclayton: We removed this months ago.

pkg/steps/clusterinstall/clusterinstall.go
smarterclayton: I could make the step return an error.  No reason we can't.

pkg/steps/clusterinstall/clusterinstall.go
smarterclayton: The e2e test step doesn't load any unique input from the graph that would alter the hash used to guarantee unique namespaces.

pkg/steps/clusterinstall/clusterinstall.go
smarterclayton: Done isn't used, suggest we remove it.

smarterclayton:
What isn't finished?

pkg/steps/template.go
smarterclayton: I'll add a test, but it only occurs when you don't ask for artifacts.

pkg/steps/clusterinstall/clusterinstall.go
stevekuznetsov: I thought we used it to trigger children?

smarterclayton:
I'm iterating on this and cleaning up issues related to dependencies.

/hold

I may create a test ci-operator image and set up test prow jobs running against it to let them soak.

smarterclayton:
Ok, removing hold.  Have tested in a normal e2e-aws job and an e2e-aws-upgrade job on prow (using slightly different config settings to verify).  The most significant issue was not using the parameters map to control access to env var checking - we were actually doing it wrong in one place and that caused other jobs to flake.

Added `--print-graph` to make it easier to spot graph failures - that output can be used with digraph.

PR #289, opened 2019-03-06
Add trailing slash to baserepo, some yum-utils complain

smarterclayton:
```
YumRepo Error: All mirror URLs are not using ftp, http[s] or file.
 Eg. http://rpm-repo-ci-op-zcd25n9t.svc.ci.openshift.org
Cannot find a valid baseurl for repo: origin-local-release
```

When we add `/` to the end it works. I've seen this behavior from
other libcurl based projects.


smarterclayton:
@bbguimaraes

PR #296, opened 2019-03-13
Include binary size and binary modification date in input hash

smarterclayton:
Ensures that if ci-operator definition changes (to a first approximation)
we build to a new namespace. Chose this over git commit because it catches
golang version changes and a few other things. We could add git commit
later but rebuild should be sufficient.


stevekuznetsov:
/lgtm
clever

PR #300, opened 2019-03-14
Convert secrets with specific shapes into pull secrets

smarterclayton:
Recognize mounted secrets of the standard form as pull secrets so the image
controllers and other tools can leverage them. Only sets the type.


smarterclayton:
@stevekuznetsov this will allow us to add a `--secret-dir .../pullsecret` and have it recognized.  However, I'm going to need to update release to create a secret with the appropriate keys (registry-pull-credentials-quay.io has a key `config.json` which doesn't match what Kube expects) by updating populate secrets to create a ci-operator-pull-secrets secret with the correct `.dockerconfigjson` key.

smarterclayton:
Blocks OCP jobs now that ci-operator imports the image vs just referencing.  Also, binary mode was wrong so I fixed that (caught it while debugging)

smarterclayton:
Tagging to unblock OCP, will react if there are any issues.

stevekuznetsov:
Nice, this is minimal

PR #301, opened 2019-03-20
Add prepublish raw_step config option to publish images from PRs

chancez:
Usage:

```
raw_steps:
- pre_publish_output_images_step:
    from: metering-helm-operator
    to:
      tag: metering-helm-operator
      namespace: metering-prerelease
```

This allows for testing of images before they get merged for use in manual testing or QE evaluation of in-progress features.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -218,6 +218,16 @@ type PromotionConfiguration struct {
 	AdditionalImages map[string]string `json:"additional_images,omitempty"`
 }
 
+type PrePublishImageTagConfiguration struct {
+	// Namespace identifies the namespace to which the built
+	// artifacts will be published to.
+	Namespace string `json:"namespace"`
+
+	// Name is an optional image stream name to use that
+	// contains all component tags.
+	Name string `json:"name"`
+}
+
 // StepConfiguration holds one step configuration.
 // Only one of the fields in this can be non-null.
 type StepConfiguration struct {
@@ -225,6 +235,7 @@ type StepConfiguration struct {
 	PipelineImageCacheStepConfiguration         *PipelineImageCacheStepConfiguration         `json:"pipeline_image_cache_step,omitempty"`
 	SourceStepConfiguration                     *SourceStepConfiguration                     `json:"source_step,omitempty"`
 	ProjectDirectoryImageBuildStepConfiguration *ProjectDirectoryImageBuildStepConfiguration `json:"project_directory_image_build_step,omitempty"`
+	PrePublishOutputImageTagStepConfiguration   *PrePublishOutputImageTagStepConfiguration   `json:"pre_publish_output_images_step,omitempty"`
 	RPMImageInjectionStepConfiguration          *RPMImageInjectionStepConfiguration          `json:"rpm_image_injection_step,omitempty"`
 	RPMServeStepConfiguration                   *RPMServeStepConfiguration                   `json:"rpm_serve_step,omitempty"`
 	OutputImageTagStepConfiguration             *OutputImageTagStepConfiguration             `json:"output_image_tag_step,omitempty"`
@@ -252,6 +263,13 @@ type OutputImageTagStepConfiguration struct {
 	// promoted unless explicitly targeted. Use for builds which
 	// are invoked only when testing certain parts of the repo.
 	Optional bool `json:"optional"`
+}
+
+// PrePublishOutputImageTagStepConfiguration  describes a step that
+// tags a pipeline image out from the build pipeline per pull-request build.
+type PrePublishOutputImageTagStepConfiguration struct {
+	From PipelineImageStreamTagReference `json:"from"`
+	To   PrePublishImageTagConfiguration `json:"to"`
 }
 
 // PipelineImageCacheStepConfiguration describes a
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -136,6 +136,14 @@ func FromConfig(
 			if _, ok := requiredNames[string(rawStep.OutputImageTagStepConfiguration.From)]; ok || !rawStep.OutputImageTagStepConfiguration.Optional {
 				stepLinks = append(stepLinks, step.Creates()...)
 			}
+		} else if rawStep.PrePublishOutputImageTagStepConfiguration != nil {
+			if len(jobSpec.Refs.Pulls) == 1 {
+				step = steps.PrePublishOutputImageTagStep(*rawStep.PrePublishOutputImageTagStepConfiguration, imageClient, imageClient, jobSpec)
+			} else if len(jobSpec.Refs.Pulls) > 1 {
+				log.Printf("pre_publish_output_images_step configured, but job has more than 1 pull-request, skipping")
+			} else {
+				log.Printf("pre_publish_output_images_step configured, but job has no pull-requests, skipping")
+			}
 		} else if rawStep.ReleaseImagesTagStepConfiguration != nil {
 			srcClient, err := anonymousClusterImageStreamClient(imageClient, clusterConfig, rawStep.ReleaseImagesTagStepConfiguration.Cluster)
 			if err != nil {
@@ -163,6 +171,12 @@ func FromConfig(
 
 		} else if rawStep.TestStepConfiguration != nil {
 			step = steps.TestStep(*rawStep.TestStepConfiguration, config.Resources, podClient, artifactDir, jobSpec)
+		} else {
+			return nil, nil, fmt.Errorf("unhandled rawStep %#v", rawStep)
+		}
+
+		if step == nil {
+			continue
 		}
 
 		step, ok := checkForFullyQualifiedStep(step, params)
--- a/pkg/steps/pre_publish_images.go
+++ b/pkg/steps/pre_publish_images.go
@@ -1,0 +1,155 @@
+package steps
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"strconv"
+
+	imageapi "github.com/openshift/api/image/v1"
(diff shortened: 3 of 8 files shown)


stevekuznetsov:
@openshift/developer-productivity-test-platform

pkg/steps/pre_publish_images.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		fromImage = from.Image.Name
+	}
+
+	is := &imageapi.ImageStream{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      conf.name,
+			Namespace: conf.namespace,
+		},
+		// TODO(chance): do I need this?
```
stevekuznetsov: No, this just helps people write `image: simple-name` in their PodSpec on the cluster and get auto resolution of image sha

pkg/steps/pre_publish_images.go
```diff
@@ -0,0 +1,189 @@
...
+	return nil, nil
+}
+
+func (s *prePublishOutputImageTagStep) getprePublishConfig() (prepublishConfig, error) {
+	if len(s.jobSpec.Refs.Pulls) == 0 {
+		return prepublishConfig{}, fmt.Errorf("did not find a pull request associated with job")
+	}
+	namespace := s.config.To.Namespace
+	if namespace == "" {
+		return prepublishConfig{}, fmt.Errorf("invalid prePublishOutputImageTagStep, namespace must be set")
```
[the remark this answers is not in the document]
chancez: Yes, I wasn't sure if we should do that since none of the existing steps except e2e seem to fail at step init. It would definitely be better to error before run, if we can.

pkg/api/types.go
```diff
@@ -218,13 +218,29 @@ type PromotionConfiguration struct {
...
 
+type PrePublishImageTagConfiguration struct {
+	// Namespace identifies the namespace to which the built
+	// artifacts will be published to.
+	Namespace string `json:"namespace"`
+
+	// Name is an optional image stream name to use that
+	// contains all component tags. If specified, tag is
+	// ignored.
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
chancez: It's one of name, or tag. So if you set name, you get  `$name:pr-123`, if you set tag, you get `pr-123:$tag`.

cmd/ci-operator/main.go
```diff
@@ -237,6 +238,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.DurationVar(&opt.cleanupDuration, "delete-after", opt.cleanupDuration, "If namespace exists for longer than this interval, delete the namespace. Set to zero to retain the contents. Requires the namespace TTL controller to be deployed.")
 
 	// actions to add to the graph
+	flag.BoolVar(&opt.prepublish, "pre-publish", false, "When images are finished building, pre-publish the into the pre-publish configuration.")
```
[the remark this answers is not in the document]
chancez: It is, but I remember you mentioning this would be good to add also, this makes it similar to promote. I'm glad to remove the extra flag.

pkg/api/types.go
```diff
@@ -218,13 +218,29 @@ type PromotionConfiguration struct {
...
+
+	// Tag is the ImageStreamTag tagged in for each
+	// build image's ImageStream.
+	Tag string `json:"tag"`
```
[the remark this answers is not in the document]
chancez: I was trying to follow how other stuff works; either push to a stream per PR, with a tag per image, or a stream per image, with a tag per PR.

pkg/steps/pre_publish_images.go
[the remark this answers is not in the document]
chancez: Oh, you mean in `Validate()`? I didn't realize that validation was done outside the steps package.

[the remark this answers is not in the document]
chancez: Oh, but validate isnt run for raw_steps. So are you thinking `defaults.go`?

cmd/ci-operator/main.go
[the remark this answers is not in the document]
droslean: But, if you use multiple targets and include the [images], you will be able to get the images from the temporary's namespace imagestream.

pkg/defaults/defaults.go
```diff
@@ -163,6 +168,12 @@ func FromConfig(
 
 		} else if rawStep.TestStepConfiguration != nil {
 			step = steps.TestStep(*rawStep.TestStepConfiguration, config.Resources, podClient, artifactDir, jobSpec)
+		} else {
+			return nil, nil, fmt.Errorf("unhandled rawStep %s", rawStep)
```
droslean: that will never happen

pkg/defaults/defaults.go
```diff
@@ -163,6 +168,12 @@ func FromConfig(
...
+		}
+
+		if step == nil {
```
droslean: why?

cmd/ci-operator/main.go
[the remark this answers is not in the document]
chancez: That isn't the intent here. The point is to publish to a different namespace that isn't the temporarily build namespace.

pkg/defaults/defaults.go
chancez: Because my step is only added if there's a pull-request currently.

chancez:
@droslean No, that only publishes to the build namespace. The purpose of this is to publish PR images to a separate namespace that isn't temporary.

pkg/steps/pre_publish_images.go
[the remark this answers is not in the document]
stevekuznetsov: We could validate raw steps, not sure why we don't

pkg/api/types.go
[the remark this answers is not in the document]
stevekuznetsov: Yeah but that is added complexity we are getting rid of because nobody uses it, so I think a smaller surface here is OK

pkg/defaults/defaults.go
chancez: If it will never happen, then we should use a panic to ensure that this case never happens due to programmer error.

stevekuznetsov: It's technically possible to add `[{}]` as your `raw_steps` config in CI Operator YAML, so it's a good error

stevekuznetsov:
@droslean @petr-muller any last thoughts here?

chancez:
I'll be adding tests and fixing up the comments soon btw, haven't forgotten, just sidetracked with other stuff and making sure I know what to fix first,

pkg/api/types.go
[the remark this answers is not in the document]
chancez: SGTM

pkg/steps/pre_publish_images.go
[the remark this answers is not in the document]
chancez: @stevekuznetsov I could add some beginnings of that probably. I'd be looping and validating raw steps that are of the the pre-publish type, ignoring the others.

chancez:
@stevekuznetsov Updated based on feedback. Let me know what you think. See commit messages for details.

pkg/defaults/defaults.go
```diff
@@ -136,6 +136,14 @@ func FromConfig(
 			if _, ok := requiredNames[string(rawStep.OutputImageTagStepConfiguration.From)]; ok || !rawStep.OutputImageTagStepConfiguration.Optional {
 				stepLinks = append(stepLinks, step.Creates()...)
 			}
+		} else if rawStep.PrePublishOutputImageTagStepConfiguration != nil {
```
droslean: where you pass this step? This will always be nil, right?

pkg/defaults/defaults.go
droslean: a `nil` step is not supposed to happen

pkg/defaults/defaults.go
chancez: No, it's a raw step, not in the top level config currently since it's still new and not something we want everyone to be using necessarily.

chancez: See https://github.com/chancez/ci-operator/blob/pre_publish/pkg/defaults/defaults.go#L440

pkg/defaults/defaults.go
chancez: It happens in the case I've added since the step isn't added when the job is a non-PR job.

chancez: I could move the `continue` into my branch  at line 146 I guess, but I still think we should check `if step == nil` and panic if it is in that case, as that would be a programmer error.

pkg/defaults/defaults.go
droslean: Did a quick view. I didnt notice it, thanks

pkg/steps/pre_publish_images.go
```diff
@@ -0,0 +1,155 @@
...
+
+type prepublishConfig struct {
+	namespace, name, tag string
+}
+
+func (s *prePublishOutputImageTagStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, error) {
+	return nil, nil
+}
+
+func (s *prePublishOutputImageTagStep) Run(ctx context.Context, dry bool) error {
```
stevekuznetsov: Is there an opportunity to DRY this out with the other publish step?

chancez: Definitely. I was going to revisit that. I extract what I can out into some helper functions.

chancez:
@stevekuznetsov DRYed up and squashed the commits to a cleaner slate.

droslean:
/lgtm
/hold

@stevekuznetsov @petr-muller Do you want to take a final look?

bparees:
@openshift/sig-logging this PR ultimately got reverted but if/when it comes back it seems like something you guys might benefit from for your test jobs?

PR #308, opened 2019-04-01
Update output_image_tag_step tests to test again pre-existing imagestreamtag

chancez:
- Refactored tests to be table driven for easier variations of existing test case.
- Updates test helper functions to be helpers, ensuring line numbers are associated with the original test case, not the helpers.
- Tests pre-existing image tag that matches desired output.
- Tests pre-existing image tag that does not match desired output.

### DIFF
--- a/pkg/steps/output_image_tag.go
+++ b/pkg/steps/output_image_tag.go
@@ -11,6 +11,7 @@ import (
 	"github.com/openshift/ci-operator/pkg/api"
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	coreapi "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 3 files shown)


chancez:
Please carefully look at my latest commit. I was working on updating my other PR and noticed that the existing logic was faulty because it didn't matter if the `AlreadyExists` check at https://github.com/openshift/ci-operator/blob/e72e7ecb5eb9b06b2f4bedd9faca049b7fd3d62e/pkg/steps/output_image_tag.go#L72-L76 was removed, the tests still passed. This is because the delete happens first, and in the fake client the deletion is immediate, but in reality it's asynchronous in real kube. This leads to the case where you may attempt to create the tag after deletion, and get an "AlreadyExists" error because the resource is still pending deletion. The existing code ignored these AlreadyExists errors, incorrectly, as this means the tag was never updated to the desired state.

pkg/steps/testing.go
```diff
@@ -111,6 +111,7 @@ func someStepLink(as string) api.StepLink {
 }
 
 func errorCheck(t *testing.T, message string, expected bool, err error) {
+	t.Helper()
```
petr-muller: Cool I did not know this, thanks!

pkg/steps/output_image_tag.go
```diff
@@ -69,29 +55,49 @@ func (s *outputImageTagStep) Run(ctx context.Context, dry bool) error {
...
+	// information about if it's a dry-run
+	from, err := s.istClient.ImageStreamTags(s.jobSpec.Namespace).Get(fmt.Sprintf("%s:%s", api.PipelineImageStream, s.config.From), meta.GetOptions{})
+	if err != nil {
+		return false, fmt.Errorf("could not resolve base image: %v", err)
+	}
+
+	desiredIst := s.imageStreamTag(from.Image.Name)
+	// if a tag already exists but doesn't match what we're looking for we're
+	// not done
+	return reflect.DeepEqual(ist.Tag, desiredIst.Tag), nil
```
petr-muller: Maybe we should use semantic `DeepEqual` to compare api objects? 

https://godoc.org/k8s.io/apimachinery/pkg/api/equality

chancez: Good idea

PR #309, opened 2019-04-02
when jobspec and git-ref co-exists, don't override the Refs.Pulls

droslean:
When `jobspec` and `git-ref` were specified, keep `jobspec`'s `Refs.Pulls`. 

In `Pulls` an `author` is specified whose name will be used to create rbac to give access to the author to the created namespace.


petr-muller:
@droslean have you tested this? I'm slightly concerned about something would try to use the `SHA` field of the `Pull`, confusing the rehearsal job...

cmd/ci-operator/main.go
```diff
@@ -286,7 +286,9 @@ func (o *options) Complete() error {
 		if err != nil {
 			return fmt.Errorf("failed to determine job spec: failed to resolve --git-ref: %v", err)
 		}
+		pulls := jobSpec.Refs.Pulls
```
stevekuznetsov: This is going to be surprising, no? In the rehearsal case we're going to see a Refs from --git-ref for target repo, but a Refs.Pulls from Prow for the release PR?

droslean:
/hold
Testing...

cmd/ci-operator/main.go
droslean: Exactly.

droslean:
@openshift/developer-productivity-test-platform 
I did some testing, and everything seems fine. Do we have to test any specific cases?

petr-muller:
Here are some risky areas:

- https://github.com/openshift/ci-operator/blob/master/cmd/ci-operator/main.go#L870-L874 here we'll assemble bad URLs (combining `spec.Refs.Org/Repo` which will be the target repo values with `spec.Refs.Pulls[0].Number` which will be a `openshift/release` PR number
- https://github.com/openshift/ci-operator/blob/master/cmd/ci-operator/main.go#L888-L893 ditto

But the thing I'm most concerned about is here: https://github.com/openshift/ci-operator/blob/master/pkg/steps/source.go#L106-L129

If I understand what is happening correctly, then we'll pass these mutant refs to `clonerefs` in the source step, and `clonerefs` [ends up trying to merge the SHA from the Pulls injected by us](https://github.com/kubernetes/test-infra/blob/master/prow/pod-utils/clone/clone.go#L199). That SHA of course would not exist in the target repository.

droslean:
/close

This PR seems to solve the problem in the rehearsal tool but creates other problems. We will try a different approach.

PR #311, opened 2019-04-03
Add prepublish raw_step config option to publish images from PRs

chancez:
This is the second attempt at PR #301.

The changes to output_image_tag_step in PR #308 make it change easier
to be done by refactoring the original change rather than reverting the
revert at 29e410ae06f2c721de324796b27fffb639ee8d4c.

This step makes it possible to push imagestreamtags built in a
ci-operator job to a specified image stream with a tag named
`$IMAGE_NAME:pr-$PR_NUMBER`

For example:
```
raw_steps:
- pre_publish_output_images_step:
    from: metering-helm-operator
    to:
      name: metering-helm-operator
      namespace: metering-prerelease
- pre_publish_output_images_step:
    from: metering-reporting-operator
    to:
      name: metering-reporting-operator
      namespace: metering-prerelease
```

Will publish imagestreamtags like the following:

```
NAME                                 DOCKER REF                                                                                                                                                 UPDATED
metering-helm-operator:pr-650        docker-registry.default.svc:5000/metering-prerelease/metering-helm-operator@sha256:d111590542171a220ff190b8d9cdf59854528de4545d40cfcf89e3b64ff57ea3        6 days ago
metering-helm-operator:pr-657        docker-registry.default.svc:5000/metering-prerelease/metering-helm-operator@sha256:66e0f0c6b0d5127e1a9794a976f7aae04532ee2eef28a920f234079b68832546        About an hour ago
metering-reporting-operator:pr-650   docker-registry.default.svc:5000/metering-prerelease/metering-reporting-operator@sha256:ae2e0d7df9ec6109e8fe70e42925a47f1d4fe74c6bd0c7c535bf4855e67f247f   6 days ago
```

### DIFF
--- a/pkg/steps/image_stream_util.go
+++ b/pkg/steps/image_stream_util.go
@@ -1,0 +1,98 @@
+package steps
+
+import (
+	"encoding/json"
+	"fmt"
+
+	imageapi "github.com/openshift/api/image/v1"
+	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
+	coreapi "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+)
+
+func createImageStreamWithTag(
+	isClient imageclientset.ImageStreamsGetter,
+	istClient imageclientset.ImageStreamTagsGetter,
+	is *imageapi.ImageStream,
+	ist *imageapi.ImageStreamTag,
+	dry bool,
+) error {
+	if dry {
+		isJSON, err := json.MarshalIndent(is, "", "  ")
+		if err != nil {
+			return fmt.Errorf("failed to marshal imagestream: %v", err)
+		}
+		fmt.Printf("%s\n", isJSON)
+
+		istJSON, err := json.MarshalIndent(ist, "", "  ")
+		if err != nil {
+			return fmt.Errorf("failed to marshal imagestreamtag: %v", err)
+		}
+		fmt.Printf("%s\n", istJSON)
+		return nil
+	}
+
+	_, err := isClient.ImageStreams(is.Namespace).Create(is)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return fmt.Errorf("could not create target imagestream: %v", err)
+	}
+
+	// Create if not exists, update if it does
+	if _, err := istClient.ImageStreamTags(ist.Namespace).Create(ist); err != nil {
+		if errors.IsAlreadyExists(err) {
+			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				existingIst, err := istClient.ImageStreamTags(ist.Namespace).Get(ist.Name, meta.GetOptions{})
+				if err != nil {
+					return err
+				}
+				// We don't care about the existing imagestreamtag's state, we just
+				// want it to look like the new one, so we only copy the
+				// ResourceVersion so we can update it.
+				ist.ResourceVersion = existingIst.ResourceVersion
+				if _, err = istClient.ImageStreamTags(ist.Namespace).Update(ist); err != nil {
+					return err
+				}
+				return nil
+			})
+			if err != nil {
+				return fmt.Errorf("could not update output imagestreamtag: %v", err)
+			}
+		} else {
+			return fmt.Errorf("could not create output imagestreamtag: %v", err)
+		}
+	}
+	return nil
+}
+
+func newImageStream(namespace, name string) *imageapi.ImageStream {
+	return &imageapi.ImageStream{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+}
+
+func newImageStreamTag(
+	fromNamespace, fromName, fromTag,
+	toNamespace, toName, toTag string,
+) *imageapi.ImageStreamTag {
+	return &imageapi.ImageStreamTag{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      fmt.Sprintf("%s:%s", toName, toTag),
+			Namespace: toNamespace,
+		},
+		Tag: &imageapi.TagReference{
+			ReferencePolicy: imageapi.TagReferencePolicy{
+				Type: imageapi.LocalTagReferencePolicy,
+			},
+			From: &coreapi.ObjectReference{
+				Kind:      "ImageStreamImage",
+				Name:      fmt.Sprintf("%s@%s", fromName, fromTag),
+				Namespace: fromNamespace,
+			},
+		},
+	}
+}
--- a/pkg/steps/pre_publish_images.go
+++ b/pkg/steps/pre_publish_images.go
@@ -1,0 +1,122 @@
+package steps
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strconv"
+
+	imageapi "github.com/openshift/api/image/v1"
+	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/ci-operator/pkg/api"
+)
+
+// PrePublishOutputImageTagStep will ensure that a tag exists
+// in the named ImageStream that resolves to the built
+// pipeline image
+type prePublishOutputImageTagStep struct {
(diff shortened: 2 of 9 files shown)


chancez:
cc @openshift/sig-logging Ben said logging might want to leverage this.

chancez:
@petr-muller  We need images before merges to master. This allows us to publish images to a non ephemeral namespace per build, allowing us to manually test changes as we develop them. Prowgen would need updating for this to work, yes. I was thinking we would add it as a raw step first, then we could move it into the main config and make it a "regular" step if it looks like it's working out well enough. Thoughts?

chancez:
We could make this a subsection within the output_image_tags step, or in the regular `images` step. We need the ability to specify the namespace at least, so it would have to be nested config option.

So something it could be something like:

```
images:
- dockerfile_path: Dockerfile.reporting-operator
  from: base
  to: metering-reporting-operator
  prepublish:
    name: metering-reporting-operator # defaults to the value of `to` if unset probably
    namespace: metering-prerelease
- dockerfile_path: Dockerfile.metering-operator
  from: metering-helm
  to: metering-helm-operator
  prepublish:
    name: metering-helm-operator
    namespace: metering-prerelease
```

chancez:
@stevekuznetsov I'm also curious on your thoughts about the above.

chancez:
If we didn't go that route, it seems like I might need something similar to the `imagesReadyStep` to make it possible to use something like `[prepublish]` as a target for all `[prepublish:$name:$tag]` targets.

petr-muller:
@chancez Thanks for explaining that use case to me. I'm using this opportunity to discuss the feature more than actually reviewing the code, so sorry about that. If we need to get the feature in quickly in some form, then iterate on that, tell me and I'll refocus on the actual current implementation.

This snippet: 

```
images:
- from: base
  to: metering-reporting-operator
  prepublish:
    name: metering-reporting-operator # defaults to the value of `to` if unset probably
    namespace: metering-prerelease
...
```
is what I had in mind, and I kinda like it more than the `raw_step` approach, especially with the default names. Taking that thought a bit further to pursue some simplification options - how much would we need to support non-default names? To me that does not seem that useful, maybe I'm wrong. And if we would not need that, then how much would we need publishing into multiple namespaces and not just one? In that case we could have just something like top-level:

```
prepublish:
  namespace: target
```

And taking *that* thought even further... Isn't that basically the same thing like promotion, just into a different place, done at a different time? Could we have multiple promotion destinations and allow users to `--promote=after_merge`, `--promote=before_merge` or something like that?

chancez:
@petr-muller I'm glad to go with just setting namespace at this point if we went with pulling this into an existing step. Can always extend further later.

I'm not so sure about how the best strategy for designing it in relationship to promotion. I found that the promotion code had expectation about controlling more than what I needed so I opted to design around something more minimal, so I designed based around the existing `output_image_tag_step`. 

I do need something relatively soon however. I'm trying to switch CI systems and this is the last thing our existing workloads require before I can get rid of our existing (jenkins based) CI infra for metering. But I don't want to rush this and want it to be done correctly.

stevekuznetsov:
@petr-muller I think the suggestion from me to @chancez to create a new step was to reduce the possible blast radius (which, hilariously, did not work) from the changes and allow him to commit on his own time and in his own workflow. I'm not opposed to an update to the promotion step but I didn't want him to chance (haha) messing up promotion for the rest of the org and be beholden to slower dev candences.

chancez:
If we keep with the raw step approach, instead of adding a new `[prepublish]` target that targets all prepublsh targets, i could also just add the multiple prepublish targets to the `[image]` deps.

stevekuznetsov:
Wouldn't `[images]` be the prereq for prepublish>

chancez:
Kinda. Right now, each prepublish step depends on 1 project image step, rather than the meta `[image]` step. I was suggesting `[image]` could be updated to have dependencies on each `[prepublish:$ns:$name]` target, and those themselves have dependencies on `[image:$ns:$name]` so it would avoid needing add anything to prowgen, and the `[image]` target would also push prepublish images.

chancez:
@petr-muller @stevekuznetsov What route would we like to go?

chancez:
The latest commit updates the PR to add a `prepublish` top-level config option and a `[prepublish]` target. The behavior is similar to the `images`/`[images]` config/target. 

You can add a config like this to your ci-operator config:
```
prepublish:
  namespace: metering-prerelease
```

Which is the same as manually writing this for each image you have:
```
raw_steps:
- pre_publish_output_images_step:
    from: metering-helm-operator
    to:
      name: metering-helm-operator
      namespace: metering-prerelease
- pre_publish_output_images_step:
    from: metering-reporting-operator
    to:
      name: metering-reporting-operator
      namespace: metering-prerelease
```

stevekuznetsov:
Direction here LGTM -- @petr-muller @bbguimaraes @droslean PTAL implemenetation

pkg/steps/pre_publish_images.go
```diff
@@ -0,0 +1,121 @@
+package steps
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strconv"
+
+	imageapi "github.com/openshift/api/image/v1"
+	"github.com/openshift/ci-operator/pkg/api"
```
petr-muller: Can you please separate local imports from others?

pkg/steps/pre_publish_images.go
```diff
@@ -0,0 +1,121 @@
...
+	return newImageStreamTag(
+		s.jobSpec.Namespace, api.PipelineImageStream, fromImage,
+		s.config.To.Namespace, s.config.To.Name, s.tag,
+	)
+}
+
+func PrePublishOutputImageTagStep(config api.PrePublishOutputImageTagStepConfiguration, istClient imageclientset.ImageStreamTagsGetter, isClient imageclientset.ImageStreamsGetter, jobSpec *api.JobSpec) api.Step {
+	pull := jobSpec.Refs.Pulls[0]
+	pullNumber := strconv.Itoa(pull.Number)
+	tag := "pr-" + pullNumber
```
petr-muller: This will be confusing if a job with prepublish is ever run by our rehearse tool. Those jobs will have a `pull` that correspond to a PR to the `openshift/release` repo, but that information is lost here, under assumption that the org/repo here will always be the "home" repo. In the worst (but rare) case you could feed images from two PRs (one to the home repo and one to openshift/release) into the same tag. To be on the safe side we could incorporate the org-repo parts to the tag name.

Rehearsing jobs with side effects is tricky anyway - the tool was built under an assumption that presubmits do not have side effects. We'll need to think about that interaction a bit.

pkg/steps/image_stream_util.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil
+	}
+
+	_, err := isClient.ImageStreams(is.Namespace).Get(is.Name, meta.GetOptions{})
+	if errors.IsNotFound(err) {
+		_, err = isClient.ImageStreams(is.Namespace).Create(is)
+	}
+	if err != nil {
+		return fmt.Errorf("could not retrieve target imagestream: %v", err)
+	}
```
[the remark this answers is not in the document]
chancez: I wondered this myself, but this logic is what's always existed. The thing to note is, if there's a conflict for the imagestream, it's fine, since the imagestream is just a container of image stream tags, so in the end, as long as it exists; we're happy (that's not true for tags however, for tags the value matters).

pkg/steps/pre_publish_images.go
chancez: True. How does this get handled by other image oriented steps? Do we need to expose the fact that something is a rehearse, and allow steps to inspect that from the jobSpec so the step can skip doing certain things under rehearse?

pkg/steps/pre_publish_images.go
chancez: Oops yes

pkg/steps/image_stream_util.go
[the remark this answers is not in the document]
petr-muller: Yeah, that's exactly why I would expect `Create` and ignore `AlreadyExists`. The `Get` followed by `Create` looks risky...

@stevekuznetsov WDYT?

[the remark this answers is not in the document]
stevekuznetsov: Yeah why not just Create and ignore the AlreadyExists? I know you copy pasta this but we might as well fix the code. I probably wrote it poorly a year ago :)

pkg/steps/pre_publish_images.go
stevekuznetsov: Slack convo said we should ignore this for now, right?

petr-muller: IMO yes

pkg/steps/image_stream_util.go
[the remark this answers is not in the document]
chancez: I can definitely change it.

PR #312, opened 2019-04-03
Make input hash independent on input ordering

petr-muller:
The hash function we use is sensitive to ordering of the string slice we
give it as input. It seems we may construct the input slice with
different ordering, so it was possible for ci-operator to place
workloads with identical inputs into different namespaces.

/cc @stevekuznetsov @smarterclayton @droslean @bbguimaraes


bbguimaraes:
@openshift-ci-robot: nailed it.

PR #313, opened 2019-04-04
Allow disabling promotion step via config

stevekuznetsov:

/assign @smarterclayton 
/cc @openshift/developer-productivity-test-platform


stevekuznetsov:
```
could not wait for build: the build src failed after 0s with reason BuildPodExists: The pod for this build already exists and is older than the build.
```

WTF?

Issue #314, opened 2019-04-04
build pod already exists and is older than the build


stevekuznetsov:
Just hit this on my PR https://github.com/openshift/ci-operator/pull/313 -- the _first_ round of tests had every single test fail with:

```
2019/04/04 03:16:59 Building src
2019/04/04 03:17:00 Build src failed, printing logs:
2019/04/04 03:17:00 error: Unable to retrieve logs from failed build: build src is in an error state. No logs are available.
2019/04/04 03:17:11 Ran for 12s
error: could not run steps: could not wait for build: the build src failed after 0s with reason BuildPodExists: The pod for this build already exists and is older than the build.
```

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_ci-operator/313/pull-ci-openshift-ci-operator-master-unit/476
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_ci-operator/313/pull-ci-openshift-ci-operator-master-lint/473
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_ci-operator/313/pull-ci-openshift-ci-operator-master-images/479

@openshift/developer-productivity-test-platform 

I've never seen this before, has anyone else? @wking

stevekuznetsov:
Ah, one of the jobs has

```
2019/04/04 03:17:00 Building src
2019/04/04 03:17:00 Build src previously failed from an infrastructure error (CannotRetrieveServiceAccount), retrying...
2019/04/04 03:17:00 Build src failed, printing logs:
2019/04/04 03:17:00 error: Unable to retrieve logs from failed build: build src is in an error state. No logs are available.
2019/04/04 03:17:11 Ran for 12s
error: could not run steps: could not wait for build: the build src failed after 0s with reason BuildPodExists: The pod for this build already exists and is older than the build.
```

stevekuznetsov:
So it would seem that our build retry logic is bad

PR #315, opened 2019-04-04
Improve build retry by ensuring cleanup succeeds

stevekuznetsov:
When removing a build with orphan or background deletion for the
associated Pod, it is possible to race with the garbage collector on the
re-creation of the Build at which point the new Build notices the old
Pod and errors out as the old Pod finished before the Build started.

/assign @smarterclayton @openshift/developer-productivity-test-platform 

If the `Delete()` call is blocking we can remove the backoff.

fixes https://github.com/openshift/ci-operator/issues/314

### DIFF
--- a/pkg/steps/source.go
+++ b/pkg/steps/source.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 
 	buildapi "github.com/openshift/api/build/v1"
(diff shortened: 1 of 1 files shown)


pkg/steps/source.go
```diff
@@ -243,10 +244,31 @@ func handleBuild(buildClient BuildClient, build *buildapi.Build, dry bool, artif
 
 			if isInfraReason(b.Status.Reason) {
 				log.Printf("Build %s previously failed from an infrastructure error (%s), retrying...\n", b.Name, b.Status.Reason)
-				opts := &meta.DeleteOptions{Preconditions: &meta.Preconditions{UID: &b.UID}}
+				zero := int64(0)
+				foreground := meta.DeletePropagationForeground
```
smarterclayton: foreground should not return until you've deleted... but retry is fine.

stevekuznetsov: Just gotta be sure yaknow?

PR #317, opened 2019-04-08
Look for likely infrastructure errors in the log snippet

stevekuznetsov:

Example snippet from one of the recent failed builds:

```
  logSnippet: "--> RUN make build WHAT=cmd/hypershift\nhack/build-go.sh cmd/hypershift
    \n++ Building go targets for linux/amd64: cmd/hypershift\n[INFO] hack/build-go.sh
    exited with code 0 after 00h 00m 48s\nerror: build error: no such image"
```

/assign @smarterclayton 
/cc @openshift/developer-productivity-test-platform

### DIFF
--- a/pkg/steps/source.go
+++ b/pkg/steps/source.go
@@ -245,7 +245,7 @@ func handleBuild(buildClient BuildClient, build *buildapi.Build, dry bool, artif
 				return fmt.Errorf("could not get build %s: %v", build.Name, err)
 			}
 
-			if isInfraReason(b.Status.Reason) {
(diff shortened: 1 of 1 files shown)


pkg/steps/source.go
```diff
@@ -314,6 +314,10 @@ func isInfraReason(reason buildapi.StatusReason) bool {
 	return false
 }
 
+func hintsAtInfraReason(logSnippet string) bool {
+	return strings.Contains(logSnippet, "no such image")
```
bparees: `error: build error: no such image` would be more precise/less likely to catch a false positive

stevekuznetsov:
@bparees addressed your comment

PR #319, opened 2019-04-10
only retry terminated builds

bparees:


pkg/steps/source.go
```diff
@@ -242,7 +242,10 @@ func handleBuild(buildClient BuildClient, build *buildapi.Build, dry bool, artif
 				return fmt.Errorf("could not get build %s: %v", build.Name, err)
 			}
 
-			if isInfraReason(b.Status.Reason) || hintsAtInfraReason(b.Status.LogSnippet) {
```
stevekuznetsov: Can you make a helper `isTerminated` for readability here

pkg/steps/source.go
```diff
@@ -301,7 +304,6 @@ func isInfraReason(reason buildapi.StatusReason) bool {
 		buildapi.StatusReasonNoBuildContainerStatus,
 		buildapi.StatusReasonFailedContainer,
 		buildapi.StatusReasonOutOfMemoryKilled,
-		buildapi.StatusReasonCannotRetrieveServiceAccount,
```
stevekuznetsov: Why?

pkg/steps/source.go
bparees: only under protest.  one exists in origin but you don't have a dep on origin.

pkg/steps/source.go
bparees: it's not a fatal condition.  the controller will keep retrying the build until the serviceaccounts are created.

bparees: I suppose technically the build could possibly timeout and get marked failed if the SAs never get created, in which case *maybe* this reason would still be there, i'm not sure.  if you want me to leave it, i can, but w/ the other changes in this PR you should not be seeing this exposed anymore anyway.

pkg/steps/source.go
bparees: updated.

stevekuznetsov: I remember making an issue for you guys to expose your helpers in an importable way like a year ago ;)

bparees: what's not importable about origin, hmmmmmm?

PR #328, opened 2019-04-12
pkg/steps: Extract image stream creation logic into helper functions

chancez:
Since we ended up with some refactoring that touches the output image tag step, in the prepublish PR, I decided to extract that code into it's own commit/PR so that we can merge that independently, and if all goes well, then the next PR related to prepublishing will not touch anything except it's own code, hopefully isolating changes better in case of issues.

### DIFF
--- a/pkg/steps/image_stream_util.go
+++ b/pkg/steps/image_stream_util.go
@@ -1,0 +1,77 @@
+package steps
+
+import (
+	"encoding/json"
+	"fmt"
+
+	imageapi "github.com/openshift/api/image/v1"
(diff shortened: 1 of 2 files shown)


chancez:
/hold
Found an issue in output_image_tag that will conflict here.

chancez:
/hold cancel
Rebased on the other change.

chancez:
@petr-muller @stevekuznetsov this is ready for review

smarterclayton:
/hold

Given scope of changes and limited review bandwidth I’m going to pause feature additions in CI-operator until we have some time to put better tests in place and have more time to review.

PR #332, opened 2019-04-17
Attempt to save current namespace resources to artifacts during exit

smarterclayton:
In order to better debug failure modes involving infrastructure, have
the ci-operator attempt to capture what was occurring at the time it
exited in the cluster. This also allows to audit when the operator isn't
giving us the results we expect, such as when it fails to properly
publish or promote images.

@stevekuznetsov as discussed earlier this will dump imagestreams, pods,
builds, templateinstances, and events to artifacts/build-resources/*.json
on exit.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -41,9 +41,11 @@ import (
 	imageapi "github.com/openshift/api/image/v1"
 	projectapi "github.com/openshift/api/project/v1"
 	templateapi "github.com/openshift/api/template/v1"
+	buildclientset "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1"
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
-	"github.com/openshift/client-go/project/clientset/versioned"
+	projectclientset "github.com/openshift/client-go/project/clientset/versioned"
 	templatescheme "github.com/openshift/client-go/template/clientset/versioned/scheme"
+	templateclientset "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
 
 	"github.com/openshift/ci-operator/pkg/api"
 	"github.com/openshift/ci-operator/pkg/interrupt"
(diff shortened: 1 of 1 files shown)


smarterclayton:
@stevekuznetsov if you have opinions here

cmd/ci-operator/main.go
```diff
@@ -864,6 +866,50 @@ func inputHash(inputs api.InputDefinition) string {
...
+
+	namespaceDir := filepath.Join(o.artifactDir, "build-resources")
+	if err := os.Mkdir(namespaceDir, 0777); err != nil {
+		log.Printf("Unable to create build-resources directory: %v", err)
+		return
+	}
+
+	if kubeClient, err := coreclientset.NewForConfig(o.clusterConfig); err == nil {
+		pods, _ := kubeClient.Pods(o.namespace).List(meta.ListOptions{})
+		data, _ := json.MarshalIndent(pods, "", "  ")
```
stevekuznetsov: Shouldn't we technically be using the Serializer here?

cmd/ci-operator/main.go
```diff
@@ -864,6 +866,50 @@ func inputHash(inputs api.InputDefinition) string {
...
+		}
+		imagestreams, _ := imageClient.ImageStreams(o.namespace).List(meta.ListOptions{})
+		data, _ := json.MarshalIndent(imagestreams, "", "  ")
+		ioutil.WriteFile(filepath.Join(namespaceDir, "imagestreams.json"), data, 0644)
+	}
+
+	if templateClient, err := templateclientset.NewForConfig(o.clusterConfig); err == nil {
+		templateInstances, _ := templateClient.TemplateInstances(o.namespace).List(meta.ListOptions{})
+		data, _ := json.MarshalIndent(templateInstances, "", "  ")
+		ioutil.WriteFile(filepath.Join(namespaceDir, "templateinstances.json"), data, 0644)
```
stevekuznetsov: This can't leak any secrets, can it?

smarterclayton: Only if we put secrets in there, which we don't today.

cmd/ci-operator/main.go
smarterclayton: It's not required, serializer is just a thin wrapper.  We require that our objects decode correctly for other reasons

PR #337, opened 2019-04-30
Use go modules

bbguimaraes:
02010da has only deletions — I suppose `go mod` is more selective about what it vendors than `glide`:

```sh
$ git diff --shortstat 02010da..02010da -- `find vendor/ -mindepth 1 -a ! -name modules.txt`
 3700 files changed, 1723593 deletions(-)
```

34e1857 is the result of `go mod tidy && go mod vendor`. Compilation and testing succeed in both revisions.


droslean:
@bbguimaraes did you test it?

stevekuznetsov:
/lgtm
/approve

If it compiles and passes our testing I imagine it will be fine

PR #338, opened 2019-05-01
When --template is specified, ignore whatever test with that name is defined

smarterclayton:
As we move templates into the ci-operator, we need a way to stage the
changes for sets of jobs and move gradually. When a user provides
--template on the command line, ignore any job with the same name so that
we can begin moving those to internal definitions.

How would we move jobs over to built in steps:

1. Land this PR to allow templates to override build in steps
2. Identify a set of ProwJobs that can omit the --template parameter that are not generated
3. Update the cluster install e2e step to match the latest template
4. Change the ProwJobs to omit the template parameter
5. Soak for several days
6. Identify any gaps in how we implement the template steps, fix ci-operator (secrets, profiles)
7. Soak more
8. Change ci-operator-prowgen to omit the --template parameter for all generated jobs
9. React and validate

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -6,10 +6,8 @@ import (
 	"encoding/base32"
 	"encoding/json"
 	"encoding/xml"
-	"errors"
 	"flag"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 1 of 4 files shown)


smarterclayton:
@petr-muller @bbguimaraes 

In order for us to start trying to move templates inline, we need a way to keep backcompat with existing jobs (which use --template) and then allow us to test a few using the version coming out of this binary, then migrate the rest by regenerating ci-operator prowgen.  This PR allows us to let `--template` override any implicit step.

PR #339, opened 2019-05-02
pkg/steps/clusterinstall/template: Update to pull in openshift/release changes

wking:
Catching up with the state as of openshift/release@f5fe3f589c741f836a30a01810b835613b8a655c (2019-05-01).

I've mostly used the release repo's `cluster-launch-installer-e2e.yaml`, but I mixed in some code from `cluster-launch-installer-openstack-e2e.yaml` because the template here seemed to have OpenStack support already.

Besides just staying current, the motivation for this is to drop from the installer's us-east-1 default of six availability zones down to two to reduce our exposure to AWS throttling for per-subnet API calls.

### DIFF
--- a/pkg/steps/clusterinstall/template.go
+++ b/pkg/steps/clusterinstall/template.go
@@ -21,6 +21,9 @@ parameters:
   required: true
 - name: RELEASE_IMAGE_LATEST
   required: true
+- name: BASE_DOMAIN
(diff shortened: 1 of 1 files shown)


wking:
Ah, @smarterclayton wanted three zones; fixed with 0bacebc -> 96f2bae.

PR #340, opened 2019-05-02
Bug 1705243: Two PR jobs can race creating an image stream tag

smarterclayton:
```
2019/05/01 14:42:27 Tagging cli into stable
...
2019/05/01 14:42:33 Ran for 15s
error: could not run steps: step [release:latest] failed: no 'cli' image was tagged into the stable stream, that image is required for building a release
```

If two jobs race, the first creates stable then continues - the second deletes the tag, but then first tries to use the tag and fails.

Replace delete then create with update image stream tag, retry on
conflict, and if it doesn't exist (or another job races with us and
it does exist), ignore that error.

There is a brief window where if the source tag is external the image
import controller might still be importing the image, but most code
in ci-operator now at least checks that condition. If it comes up we
can have output image stream tag loop and wait for import like the
release command does.

### DIFF
--- a/pkg/steps/output_image_tag.go
+++ b/pkg/steps/output_image_tag.go
@@ -7,6 +7,8 @@ import (
 	"log"
 	"strings"
 
+	"k8s.io/client-go/util/retry"
+
 	imageapi "github.com/openshift/api/image/v1"
 	"github.com/openshift/ci-operator/pkg/api"
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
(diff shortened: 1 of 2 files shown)


pkg/steps/output_image_tag.go
```diff
@@ -54,16 +56,16 @@ func (s *outputImageTagStep) Run(ctx context.Context, dry bool) error {
...
-	}
-
-	// Create if not exists, if it already exists, then we have nothing to do.
-	if _, err := s.istClient.ImageStreamTags(toNamespace).Create(ist); err != nil && !errors.IsAlreadyExists(err) {
-		return fmt.Errorf("could not create output imagestreamtag: %v", err)
+	// ensure that the image stream tag points to the correct input, retry
+	// on conflict, and do nothing if another user creates before us
+	if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		_, err := s.istClient.ImageStreamTags(toNamespace).Update(ist)
+		if errors.IsNotFound(err) {
```
droslean: How often the image already exists? Do you think that it's better first to try to create the image and if its already exist then update?

droslean:
I left a nit comment that can be done in a follow-up. It's better to land this first to resolve the bug.
/lgtm

pkg/steps/output_image_tag.go
smarterclayton: the image will almost always exist (we create stable based on the release tags, then we modify it)

smarterclayton:
This looks fixed, no recurrences in 24h 

https://search.svc.ci.openshift.org/?search=image+was+tagged+into+the+stable+stream&maxAge=168h&context=2&type=all

PR #344, opened 2019-05-10
 Add an option to report failures to Sentry

petr-muller:
Add a `--sentry-dsn-path` parameter pointing to a file containing the
Sentry DSN secret. If provided, ci-operator will report failures to
Sentry using this secret. Initially, the reports are message-based
(see https://docs.sentry.io/data-management/rollups for more
information about how Sentry groups events), which means only
errors with exactly matching error messages will be grouped together
in Sentry. The other option was to use pseudo-exceptions made from
the `error`: that causes the other extreme, **all** events would be
coalesced into a single issue (because our errors do not carry the
original stack trace, so we would always report an error with an identical
stack trace). Sentry can find "similar" issues based on message similarity,
this will be a good source for finding out how to improve our error
messaging/reporting so that errors are better grouped in Sentry.

The errors are annotated by a basic set of tags read from `JOB_SPEC`.

Decided to use a third-party https://github.com/SierraSoftworks/sentry-go
package as a Sentry client. We will likely need the fingerprinting
features to break down our failure events and the official client does
not support this Sentry feature. Plus, IMHO the API is better.

/cc @stevekuznetsov @droslean @bbguimaraes @hongkailiu


stevekuznetsov:
If we pipe through `logrus` entries we can use e.g. [caller](https://github.com/sirupsen/logrus#logging-method-name) metadata to pinpoint the file name and method instead of error text? We get a stack as well IIRC.

cmd/ci-operator/main.go
```diff
@@ -910,6 +916,68 @@ func (o *options) saveNamespaceArtifacts() {
...
+	tags := map[string]string{
+		"prowjob-type": string(jobSpec.Type),
+		"job":          jobSpec.Job,
+		"build-id":     jobSpec.BuildId,
+		"prowjob-id":   jobSpec.ProwJobID,
+	}
+
+	userInfo := sentry.UserInfo{ID: "ci-operator", Username: "ci-operator"}
+
+	if jobSpec.Refs != nil {
```
stevekuznetsov: Could make this generic for ExtraRefs too?

stevekuznetsov:
Let's merge this as-is

/lgtm

cmd/ci-operator/main.go
```diff
@@ -465,6 +469,7 @@ func (o *options) Run() error {
 		}
 		if err != nil {
 			if !o.dry {
+				o.reportToSentry(err)
```
bbguimaraes: I was expecting to see this in the outermost scope. Don't we miss important failures by reporting only `steps.Run`?

cmd/ci-operator/main.go
```diff
@@ -910,6 +915,32 @@ func (o *options) saveNamespaceArtifacts() {
...
+func (o *options) reportToSentry(toReport error) {
+	if o.sentryDSNPath == "" || toReport == nil {
+		return
+	}
+
+	rawDsn, err := ioutil.ReadFile(o.sentryDSNPath)
+	if err != nil {
+		log.Printf("Failed to read Sentry DSN from %s: %v", o.sentryDSNPath, err)
+		return
+	}
```
bbguimaraes: This should be done at the beginning of the execution.

bbguimaraes: #3 is what I had in mind, but your justification makes sense.

cmd/ci-operator/main.go
petr-muller: Not that many, but you are right - I got lured there by an already existing dry run conditional block (I think we only want to report in real runs). I'll file a followup.

wking:
> Decided to use a third-party https://github.com/SierraSoftworks/sentry-go package as a Sentry client. We will likely need the fingerprinting features to break down our failure events and the official client does not support this Sentry feature.

No?  There's  a `Fingerprint` property on [`Packet`][1].

[1]: https://godoc.org/github.com/getsentry/raven-go#Packet

petr-muller:
> No? There's a Fingerprint property on Packet.

Right, I was not 100% precise. Not unsupported, but not well-supported. Working with `Packet`s is kinda low-level - there is no method which would create reasonably populated `Packet` for some situation (from a err instance) - you either use `CaptureX` methods which assemble `Packet` for you but you cannot modify it (add the fingerprint), or you use `Capture` which takes `Packet` as a param, but you need to assemble it yourself.

PR #347, opened 2019-05-15
Remove fake option validation code

petr-muller:
`Validate` is a NOP, so let's get rid of some code. It can be easily added once we want to validate something.

/cc @stevekuznetsov @bbguimaraes @droslean @hongkailiu


petr-muller:
LOL we even had a test for it :D

stevekuznetsov:
What on earth am I looking at. Why not instead of this, add actual tests and validation for the command line options?

petr-muller:
> What on earth am I looking at. Why not instead of this, add actual tests and validation for the command line options?

Not having validation was clearly not a problem for ~10 months, so adding it does not seem to be a priority. Removing a bunch of useless (plus confusing, because at `main()` level you are falsely assuming there *is* some validation going on) code  is a 1m operation.

PR #348, opened 2019-05-17
Make artifacts container survive longer after signal

petr-muller:
Artifacts container lives for a while after a first ci-operator instance
manages to fetch the artifacts from it. There may be *other* ci-operator
instance waiting on the pod and fetching artifacts, though, so let them
have more time to finish their artifact grab.

This is not a proper fix of the race, just a bandaid to reduce
occurrences in production. I can watch if the bandaid has a desired effect
in [Sentry](https://sentry.io/organizations/red-hat/issues/1024995691/?project=1456253)

See: https://bugzilla.redhat.com/show_bug.cgi?id=1693865#c12

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -316,7 +316,9 @@ touch /tmp/done
 echo "Waiting for artifacts to be extracted"
 while true; do
 	if [[ ! -f /tmp/done ]]; then
-		echo "Artifacts extracted"
+		echo "Artifacts extracted, will terminate after 30s"
+		sleep 30
+		echo "Exiting"
 		exit 0
 	fi
 	sleep 5 & wait


pkg/steps/artifacts.go
```diff
@@ -316,7 +316,9 @@ touch /tmp/done
 echo "Waiting for artifacts to be extracted"
 while true; do
 	if [[ ! -f /tmp/done ]]; then
-		echo "Artifacts extracted"
+		echo "Artifacts extracted, will terminate after 60s"
```
smarterclayton: Why do you say 60s here?

smarterclayton: If you want to delay shutdown, it would be better to have a prestop hook.

petr-muller: b/c I'm an idiot, had `sleep 60` here first, then shortened it to 30s and forgot to fix the message :)

I'll look into prestop hooks (I'm not familiar with them) if you want me to, but this is really just a band-aid so that we don't hit the problem that often.

I'd like to eventually fix this properly by implementing some synchronization so that the artifacts container never terminater before all ci-operator instances that wait for the pod and extract artifacts finish.

petr-muller:
/hold 

For some reason the variant with the hook does not work for me as expected, the container simply terminates. Am I reading the [docs](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks) correctly that the hooks are actually called **only** when the pods needs to be externally stopped? Because in this case the container just terminates by the process simply exiting.

petr-muller:
/hold cancel

Reverting back to the "just sleep inside the process" commit. I found hooks to not execute when the container simply exits, and I confirmed this is the case on [#forum-pod](https://coreos.slack.com/archives/CDDEUQV8X/p1558373988044100), so hooks are not applicable here.

PR #349, opened 2019-05-18
Allow the upgrade template to accept options as env

smarterclayton:
In order to perform disruptive upgrade tests, the commands for
a job need to be able to set TEST_OPTIONS to the arg on openshift-tests.

The rollback test will pass `abort-at=random` and the disruptive test will
trigger master reboots periodically.


smarterclayton:
Tagging so I can begin testing this, @petr-muller we should probably set an owners file for this repo that is more the set of people who own templates vs ci-operator itself (will be more of an issue as we have more templates in here).

PR #353, opened 2019-05-27
template: wait for pod to teardown (if container is present) during delete

vrutkovs:
Pod teardown may take longer than 5 mins (default ci-operator timeout). 
This commit would ensure the same timeout is applied to the teardown 
container - and then applied to the pod again.

This is useful for rehearse jobs, which reuse the namespace when testing 
a new commit


pkg/steps/template.go
```diff
@@ -434,6 +434,14 @@ func waitForPodDeletion(podClient coreclientset.PodInterface, name string, uid t
 		if pod.UID != uid {
 			return nil
 		}
+		for j := 0; j < timeout; j += 2 {
```
petr-muller: This does not really look right to me... This cycle will call `podRunningContainers(pod)` 150 times in immediate succession on each iteration of the outer cycle. which seems excessive and not useful. Also, isn't `pod` invariant here?

vrutkovs: >This cycle will call podRunningContainers(pod) 150 times in immediate succession on each iteration of the outer cycle

Oops, sorry, added `time.Sleep(2 * time.Second)` to wait for teardown container to complete.

This still doesn't look optimal to me though. I wonder if we can setup a watch for pods instead of fetching pod details every time?

petr-muller: Yeah, two nested wait loops do not seem that optimal to me either - worst case, this could wait what, 2*150*300 seconds? That's more than a day... 

But still, does that cycle even make sense? To me this looks like `podRunningContainers(pod)` would return the same result on each iteration, `podRunningContainers` seems to work with just a Go object - it won't change between the nested cycle iterations. 

This will need something smarter, imho, just like you suggest.

vrutkovs: Ah, right, `podRunningContainers` includes Terminated

vrutkovs:
Created a better version of this:
1) added a few helper functions
2) Before the loop wait for teardown to complete (if this container exists in the pod)
3) Wait for pod to be deleted

I didn't get a chance to test this yet

pkg/steps/template.go
```diff
@@ -421,18 +421,50 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+	if pod.UID != uid {
+		return nil, nil
+	}
+	return pod, nil
+}
+
+func waitForPodDeletion(podClient coreclientset.PodInterface, name string, uid types.UID) error {
+	timeout := 300
+	pod, err := getPodByName(podClient, name, uid)
+	if pod != nil {
```
petr-muller: Uh, `err != nil`?

pkg/steps/template.go
```diff
@@ -421,18 +421,50 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+	// Attempts to wait for teardown to complete
+	containerNames := podContainerNames(pod)
+	if sort.SearchStrings(containerNames, "teardown") < len(containerNames) {
+		waitForTeardownToComplete(pod)
+	}
+
+	// Wait for pod to be completed
+	for i := 0; i < timeout; i += 2 {
+		pod, err := getPodByName(podClient, name, uid)
+		if pod != nil {
```
petr-muller: `err != nil`

pkg/steps/template.go
```diff
@@ -421,18 +421,50 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
 		}
-		if pod.UID != uid {
-			return nil
+	}
+}
+
+func getPodByName(podClient coreclientset.PodInterface, name string, uid types.UID) (*coreapi.Pod, error) {
+	pod, err := podClient.Get(name, meta.GetOptions{})
+	if errors.IsNotFound(err) {
+		return nil, nil
```
petr-muller: I think returning `nil, nil` from a function called `getPodByName` when the pod is not found is quite confusing.

pkg/steps/template.go
```diff
@@ -421,18 +421,50 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+func waitForTeardownToComplete(pod *coreapi.Pod) {
 	timeout := 300
 	for i := 0; i < timeout; i += 2 {
-		pod, err := podClient.Get(name, meta.GetOptions{})
-		if errors.IsNotFound(err) {
-			return nil
-		}
-		if err != nil {
-			return fmt.Errorf("could not retrieve deleting pod: %v", err)
+		if !podTeardownCompleted(pod) {
```
petr-muller: I pointed this out in the previous reviews, so I'll try to elaborate. I might be wrong, but I think we get `pod` as a param and it does not change at all while `waitForTeardownToComplete` executes (there is no API call that would update the pod status). So either container statuses in `pod` are from the start set that we would break on the first iteration, or they force us to wait, but `pod` won't change while we sleep, so we would always sleep the full 300s.

pkg/steps/template.go
vrutkovs: nope, we have `return nil` in some cases (e.g. wrong UID) when pod is gone

pkg/steps/template.go
vrutkovs: Yeah, name is confusing, maybe `checkThatPodExists`?

pkg/steps/template.go
vrutkovs: ah, right, seems `waitForTeardownToComplete` should run `getPodByName` first, check that `teardown` container is present and then run the loop

pkg/steps/template.go
petr-muller: yeah, but in the happy path we would get `pod, nil`, and would return instead of waiting for deletion.

vrutkovs: Reworked this in 93f52a7

pkg/steps/template.go
vrutkovs: Renamed in 2852853

pkg/steps/template.go
vrutkovs: Reworked in c4d665d - now all functions are attempting to fetch pod info

vrutkovs:
>what problem does this actually solve? If we wait for a pod to be deleted, what good is waiting for its containers to terminate?

ci-operator would wait for 300 secs only. If teardown didn't finish by that time the pod would be removed: leftover artifacts (usually Route53 records) would remain and cause issues on next retest

pkg/steps/template.go
```diff
@@ -421,18 +421,57 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
-			return fmt.Errorf("could not retrieve deleting pod: %v", err)
+			return err
 		}
-		if pod.UID != uid {
-			return nil
+	}
+
+	// Wait for pod to be completed
+	for i := 0; i < timeout; i += 2 {
+		if _, err := checkPodExists(podClient, name, uid); err != nil {
```
petr-muller: This is getting tricky, now we never end successfully.

vrutkovs: It would throw `errors.NewGone("pod was removed")` and immediately return if the pods is gone, isn't it?

petr-muller: Well yeah, but previously ` waitForPodDeletion` returned `nil` when the pod was "successfully" not found, so the callsites probably rely on that.

vrutkovs: Ah, agree, `abb16d4` fixes that - `waitForPodDeletion` now returns nil in this case

vrutkovs:
/hold

I can't come up with a way to test this yet

pkg/steps/template.go
```diff
@@ -421,17 +421,57 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+			break
 		}
+	}
+	return nil
+}
+
+func checkPodExists(podClient coreclientset.PodInterface, name string, uid types.UID) (*coreapi.Pod, error) {
+	pod, err := podClient.Get(name, meta.GetOptions{})
+	if errors.IsNotFound(err) {
+		return nil, errors.NewGone("pod was removed")
```
stevekuznetsov: notfound is expected in the waitForTeardownToComplete -- this should not be an error?

pkg/steps/template.go
```diff
@@ -421,17 +421,57 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+	}
+	if err != nil {
+		return nil, fmt.Errorf("could not retrieve deleting pod: %v", err)
+	}
+	if pod.UID != uid {
+		return nil, errors.NewBadRequest("mismatching UID")
```
stevekuznetsov: Similarly ... a new pod with the same name but a different UUID just means teardown did complete?

pkg/steps/template.go
```diff
@@ -620,6 +660,29 @@ func templateInstanceReady(instance *templateapi.TemplateInstance) (ready bool,
...
+func podContainerNames(pod *coreapi.Pod) []string {
+	var names []string
+	for _, status := range append(append([]coreapi.ContainerStatus{}, pod.Status.InitContainerStatuses...), pod.Status.ContainerStatuses...) {
+		names = append(names, status.Name)
+	}
+	sort.Strings(names)
+	return names
+}
+
+func podTeardownCompleted(pod *coreapi.Pod) bool {
```
stevekuznetsov: this technically returns `true` when there is no `teardown` container at all ...

vrutkovs:
>Why are we making this change?

When test gets cancelled - a new commit pushed in rehearse tests for instance - ci-operator would send termination signal and wait for pod to be gone for 5 mins. In most install tests teardown + artifacts take longer than 5 mins.

This change would wait longer if pod has `teardown` container - so that it won't affect unit tests / release jobs etc. It would first wait 5 mins for teardown to complete and then 5 mins for pod to disappear (in most of the installs teardown completes in 5-7 minutes).

See also https://github.com/openshift/ci-operator/pull/353#issuecomment-498171087

pkg/steps/template.go
vrutkovs: Different UUID means the existing pod is now gone and we no longer have to wait for teardown to complete, so its expected

pkg/steps/template.go
vrutkovs: Right, if there is no teardown container there is no need to additionally wait for it to complete

pkg/steps/template.go
stevekuznetsov: But you're returning an error that you then bubble up ..?

stevekuznetsov:
OK, makes sense. Why not start a watch?

vrutkovs:
Reworked this to leverage polls and watches:
* `waitForTeardownToComplete` polls for a list of containers
* `waitForPodDeletion` watches for pod to be removed with a timeout

pkg/steps/template.go
vrutkovs: Added description to the function. In short it would return error if the pod doesn't exist or has a different UID. In `waitForTeardownToComplete` and `waitForPodDeletion` this error doesn't get propagated though

pkg/steps/template.go
```diff
@@ -421,24 +421,77 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+	}
+	return pod, nil
+}
 
-	return fmt.Errorf("waited for pod %s deletion for %ds, was not deleted", name, timeout)
+func waitForPodDeletion(podClient coreclientset.PodInterface, name string, uid types.UID) error {
+	timeout := 5 * time.Minute
+	pod, err := checkPodExistsAndValid(podClient, name, uid)
+	if err != nil {
+		// Pod with this name and UID is gone, so its successfully deleted
```
petr-muller: or, the `fmt.Errorf("could not retrieve deleting pod: %v", err)` on L452 happened, which is likely some infra error and should bubble up. 

I think it may be fine to assume that either the pod does not exist or the infra error will bring down the run a bit later, but this assumption would need to be documented here.

vrutkovs: Passing infra errors in d2eb3c now, PTAL

pkg/steps/template.go
```diff
@@ -421,24 +421,78 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+	return nil
+}
+
+// Check that pod with this name exists and has the same UID
+func checkPodExistsAndValid(podClient coreclientset.PodInterface, name string, uid types.UID) (*coreapi.Pod, error) {
+	pod, err := podClient.Get(name, meta.GetOptions{})
+	switch {
+	case err != nil:
+		return nil, fmt.Errorf("could not retrieve deleting pod: %v", err)
+	case errors.IsNotFound(err), pod.UID != uid:
```
petr-muller: the `err != nil` case will hit when `errors.IsNotFound(err)` is true, we need to check for the specific case first.

vrutkovs: Good catch, updated

petr-muller:
LGTM, let's give @stevekuznetsov a chance to review

pkg/steps/template.go
```diff
@@ -421,24 +421,78 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+				// Teardown container finished
+				return true, nil
+			}
+		}
+		return false, nil
+	})
+	return nil
+}
+
+// Check that pod with this name exists and has the same UID
```
stevekuznetsov: this is misleading

pkg/steps/template.go
```diff
@@ -421,24 +421,78 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
-			return nil
+		if pod == nil {
+			// Pod was destroyed
+			return true, nil
 		}
-		log.Printf("Waiting for pod %s to be deleted ... (%ds/%d)", name, i, timeout)
-		time.Sleep(2 * time.Second)
+
+		for _, status := range append(append([]coreapi.ContainerStatus{}, pod.Status.InitContainerStatuses...), pod.Status.ContainerStatuses...) {
+			if status.Name == "teardown" && status.State.Terminated != nil {
```
stevekuznetsov: Is `teardown` ever an `initcontainer`?

pkg/steps/template.go
```diff
@@ -421,24 +421,78 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
-	for i := 0; i < timeout; i += 2 {
-		pod, err := podClient.Get(name, meta.GetOptions{})
-		if errors.IsNotFound(err) {
-			return nil
-		}
+func waitForTeardownToComplete(podClient coreclientset.PodInterface, name string, uid types.UID) error {
+	timeout := 5 * time.Minute
+
+	log.Printf("Waiting for pod %s to complete teardown ...", name)
+	wait.Poll(10*time.Second, timeout, func() (done bool, err error) {
```
stevekuznetsov: We used to poll ever 2s -- why change?

pkg/steps/template.go
```diff
@@ -421,24 +421,78 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+	case err != nil:
+		return nil, fmt.Errorf("could not retrieve deleting pod: %v", err)
+	}
+	return pod, nil
+}
+
+func waitForPodDeletion(podClient coreclientset.PodInterface, name string, uid types.UID) error {
+	timeout := 5 * time.Minute
+	pod, err := checkPodExistsAndValid(podClient, name, uid)
+	if err != nil || pod == nil {
```
stevekuznetsov: in the case that `err == nil` but `pod ==nil` why do you return a `nil` err here? please leave a comment

pkg/steps/template.go
```diff
@@ -421,24 +421,78 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+	case errors.IsNotFound(err), pod.UID != uid:
+		// Pod was removed or mismatching UID
+		return nil, nil
+	case err != nil:
+		return nil, fmt.Errorf("could not retrieve deleting pod: %v", err)
+	}
+	return pod, nil
+}
+
+func waitForPodDeletion(podClient coreclientset.PodInterface, name string, uid types.UID) error {
```
stevekuznetsov: `waitForPodDeletion` is no longer valid as a name -- were all callers expecting this new behavior?

pkg/steps/template.go
```diff
@@ -421,24 +421,78 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+	// Attempts to wait for teardown to complete
+	containerNames := podContainerNames(pod)
+	if sort.SearchStrings(containerNames, "teardown") < len(containerNames) {
+		err = waitForTeardownToComplete(podClient, name, uid)
+		if err != nil {
+			return err
+		}
+	}
+
+	watcher, err := podClient.Watch(meta.ListOptions{
```
stevekuznetsov: If you're setting up a watch, why not just use it for all of the interaction? Why the poll?

pkg/steps/template.go
```diff
@@ -421,24 +421,78 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+		Watch:         true,
+	})
+	if err != nil {
+		return fmt.Errorf("could not create watcher for pod: %v", err)
 	}
+	defer watcher.Stop()
 
-	return fmt.Errorf("waited for pod %s deletion for %ds, was not deleted", name, timeout)
+	log.Printf("Waiting for pod %s to be deleted in %d seconds", name, timeout)
+	_, err = watch.Until(timeout, watcher, func(event watch.Event) (done bool, err error) {
```
stevekuznetsov: Why a 5min watch for deleted _after_ a 5min retry on the container step?

pkg/steps/template.go
```diff
@@ -620,6 +674,15 @@ func templateInstanceReady(instance *templateapi.TemplateInstance) (ready bool,
 	return false, nil
 }
 
+func podContainerNames(pod *coreapi.Pod) []string {
+	var names []string
+	for _, status := range append(append([]coreapi.ContainerStatus{}, pod.Status.InitContainerStatuses...), pod.Status.ContainerStatuses...) {
+		names = append(names, status.Name)
+	}
+	sort.Strings(names)
```
stevekuznetsov: Why?

pkg/steps/template.go
```diff
@@ -421,24 +421,78 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+func waitForPodDeletion(podClient coreclientset.PodInterface, name string, uid types.UID) error {
+	timeout := 5 * time.Minute
+	pod, err := checkPodExistsAndValid(podClient, name, uid)
+	if err != nil || pod == nil {
+		return err
+	}
+
+	// Attempts to wait for teardown to complete
+	containerNames := podContainerNames(pod)
+	if sort.SearchStrings(containerNames, "teardown") < len(containerNames) {
```
stevekuznetsov: nit: I like `if sets.NewString(containerNames).Has("teardown")` a lot more than these types of manipulations

pkg/steps/template.go
vrutkovs: Artifacts upload also take time to complete

pkg/steps/template.go
vrutkovs: I don't think container terminate status can be watched, can it?

stevekuznetsov: Why not? You'd get any changes to PodStatus if I understand Watches correctly

stevekuznetsov: (watch the Pod, not the container)

vrutkovs:
Simplified this:
* If pod has teardown container timeout is extended by 10 mins (average teardown takes 10 min on AWS, but might take longer on openstack)
* `waitForPodDeletion` now uses a watcher instead of polling.
  Should it also print the message when teardown has completed?

pkg/steps/template.go
```diff
@@ -422,23 +422,65 @@ func createOrRestartPod(podClient coreclientset.PodInterface, pod *coreapi.Pod)
...
+	defer watcher.Stop()
+
+	log.Printf("Waiting for pod %s to be deleted in %d seconds", name, timeout)
+	_, err = watch.Until(timeout, watcher, func(event watch.Event) (done bool, err error) {
+		switch event.Type {
+		case watch.Deleted:
+			// pod was deleted
+			return true, nil
+		case watch.Added, watch.Modified:
+			if hasTeardownContainer {
```
stevekuznetsov: I don't understand this logic. If we have a `teardown` container, we will exit out early every time. If we _don't_ have a `teardown` container, we set this boolean to `false`. The comment says that will avoid re-checking, but in reality that means we check every time. Then, if the `teardown` container is terminated, you signal you are done, so the watch ends. I think we just need a dead-simple watch, or two watches. If you want to have _one_ watch with a variable timeout, just wait for the deletion. If you want to wait for the `teardown` container completion and the pod deletion separately, you will want separate watches. 

In general, if the issue was a too-short timeout that cut the `teardown` container short, why not just make this watch go on for an hour? In what cases do we _not_ want to wait for the Pod to really be gone?

Also, if you look at the implementations in the build utils, we want a list then a watch with retires to handle transient errors.

vrutkovs: >If we have a teardown container, we will exit out early every time

Fixed by introducing `teardownFinished` var

>If you want to wait for the teardown container completion and the pod deletion separately, you will want separate watches.

That was my initial idea (see f758b7df804c1f9bdf7a796ad9a762acc284c784), however there is short time (between teardown watch and pod to be deleted watch) where pod may be destroyed and could have been replaced with a new pod. Two watches don't seem reliable to me

>why not just make this watch go on for an hour? In what cases do we not want to wait for the Pod to really be gone?

That would hide potential issues in teardown

>we want a list then a watch with retires to handle transient errors.

Using `event, ok := <-watcher.ResultChan()`? It doesn't seem to have some kind of timeout

stevekuznetsov:
> That would hide potential issues in teardown

Which? Before we spend more time working on an implementation can we determine why the (stupid simple) approach of doing more retries over a 10, 20, 30 minute period would not be appropriate? Do we have some SLA for teardown time?

vrutkovs:
Extending the timeout is the simplest approach, and its valid, however it would apply to all ci-operator pods. 

e2e-aws's teardown is the only one I know of which takes longer than 5 mins at the moment, other types of tests may rely on existing timeout.

This PR is just one possible way, of course. If it looks overcomplicated then lets just bump the timeout on teardown to fix rehearse failures at least

stevekuznetsov:
Of course it would hit all pods, but we poll every 2 seconds right now, so the only cases where increasing the timeout would actually increase the time taken for the test to run is if the pod is not returning in the current timeout, and then it would only increase it by the time taken to finish tearing down, right?

vrutkovs:
Right, it appears a larger timeout would act the same. Created #358 instead.

I'll keep this open for now

